PR #9, opened 2018-04-08
Added Xargo.toml

lachlansneff:
This allows the bootloader to compile with newer nighties.

https://github.com/rust-lang/rust/pull/49503#issuecomment-377559413

### DIFF
--- a/Xargo.toml
+++ b/Xargo.toml
@@ -1,0 +1,6 @@
+[dependencies.core]
+stage = 0
+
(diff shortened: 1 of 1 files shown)


robert-w-gries:
With latest `xargo 0.3.12`,  `compiler_builtins` is compiled by default.

See japaric/xargo#213

lachlansneff:
My apologies, I didn't see that the new version of xargo had been released.

PR #17, opened 2018-07-01
Update to new panic_implementation attribute

IsaacWoods:
This updates the old lang item to the new `#[panic_implementation]` attribute, fixing the bootloader on new nightlies

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,6 +8,7 @@
 #![feature(pointer_methods)]
 #![feature(const_fn)]
 #![feature(nll)]
+#![feature(panic_implementation)]
 #![no_std]
 #![no_main]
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -214,16 +217,15 @@ fn enable_write_protect_bit() {
...
-    _column: u32,
-) -> ! {
+pub extern "C" fn panic(info: &PanicInfo) -> ! {
     use core::fmt::Write;
-    write!(printer::Printer, "PANIC: {} in {}:{}", msg, file, line).unwrap();
+    if let Some(location) = info.location() {
+        write!(printer::Printer, "PANIC: {} in {}:{}", info.message().unwrap(), location.file(), location.line()).unwrap();
+    } else {
+        write!(printer::Printer, "PANIC: {} at ???", info.message().unwrap()).unwrap();
+    }
```
phil-opp: PanicInfo implements Display, so I think we can just do `println!("PANIC: {}", info)` or even `println!("{}", info)`. We also wouldn't need `#![feature(panic_info_message)]` then. Or is there some advantage in printing explicitly?

IsaacWoods: Nope, I just missed that `PanicInfo` implemented `Display`. Thanks for the review; fixed!

phil-opp:
Thanks!

phil-opp:
Published as v0.2.0-alpha-005

PR #19, opened 2018-07-15
Add more explanations to some opaque x86 instructions

anirudhb:
Usually the reader will have to search for the definitions of the instructions. By including the (short) description, they can continue to go through the code without needing to do a search.


phil-opp:
Thanks a lot!

Issue #14, opened 2018-05-22
Don't mask IRQs


phil-opp:
Currently the bootloader masks all hardware interrupts: https://github.com/rust-osdev/bootloader/blob/bda13d807d0c7451b1ffe74c6c97f6369aef6ff9/src/second_stage.s#L68-L71

They shouldn't be masked when control is passed to the kernel.

robert-w-gries:
Hey Phil, are you saying that interrupts should be enabled when the bootloader jumps to `_start()`?

phil-opp:
No, enabling them would not be a good idea since there isn't an IDT yet. I mean that they shouldn't be _masked_. An interrupt can be masked in the interrupt controller, so that the controller does not emit such an interrupt, even when interrupts are generally enabled.

rybot666:
Since this was patched, can we close the PR?

phil-opp:
@rybot666 Yes, thanks for the hint!

PR #20, opened 2018-07-19
Version 0.2.0

phil-opp:
Given that we were in alpha state for a long time and there were no real bugs reported, I think it's time to finally release version 0.2.0. Another big reason for doing it is that the current versions of bootimage only work with the latest alpha versions, so everybody was forced to use the alpha versions anyway.

There are a few improvements in this PR:

<ul>
<li> Add a `lib.rs` that re-exports the `os_bootinfo` crate as `bootinfo`. This works nicely with the [new version of `bootimage`](https://github.com/rust-osdev/bootimage/pull/16). It means that you no longer need to add a `os_bootinfo` dependency and instead add the `bootloader` as dependency directly. This solves the potential version mismatch problem (i.e. kernel and bootloader use different `os_bootinfo` versions).</li>
<li>Add an `entry_point` macro which allows to specify the entry point like this:

```rust
#[macro_use]
extern crate bootloader;

#[cfg(not(test))]
entry_point!(kernel_main);

fn kernel_main(boot_info: &'static bootloader::bootinfo::BootInfo) -> ! {…}
```

The advantage of using this macro is that the function signature is type-checked, i.e. it is an error if the argument type is wrong. You also no longer need `no_mangle`, `extern "C"`, etc since the macro handles this for us.
</li>
<li>
- Disable interrupts instead of masking them. This solves #14.
</li>
</ul>

This PR is **already pulished as `0.2.0-beta`**. Please test it and give feedback!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bootloader"
-version = "0.2.0-alpha-005"
+version = "0.2.0-beta"
 dependencies = [
  "fixedvec 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "os_bootinfo 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@ cargo-features = ["publish-lockfile"]
 
 [package]
 name = "bootloader"
-version = "0.2.0-alpha-005"
+version = "0.2.0-beta"
 authors = ["Philipp Oppermann <dev@phil-opp.com>"]
 license = "MIT/Apache-2.0"
 description = "An experimental pure-Rust x86 bootloader."
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,23 @@
 #![no_std]
 
 pub extern crate os_bootinfo as bootinfo;
+
+/// Defines the entry point function.
+///
+/// The function must have the signature `fn(&'static BootInfo) -> !`.
+///
+/// This macro just creates a function named `_start`, which the linker will use as the entry
+/// point. The advantage of using this macro instead of providing an own `_start` function is
+/// that the macro ensures that the function and argument types are correct.
+#[macro_export]
+macro_rules! entry_point {
+    ($path:path) => {
+        #[export_name = "_start"]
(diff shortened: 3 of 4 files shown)


phil-opp:
Having it as a separate crate allows multiple bootloader implementations to reuse the same bootinfo struct, so we might want to keep it separate. However, I completely agree that we should hide the fact that this is a common crate to be less constrained by backwards compatibility in the future.

I think the best way is to create a `bootinfo` module in the bootloader and reexport all os_bootinfo types from it. This way we can move the implementation into the bootloader without a breaking change later (if we want to). The user never sees the os_bootinfo crate.

phil-opp:
I pushed a new version that does not expose `os_bootinfo`, but instead reexports its types as part of a public `bootinfo` module. This should allow us to move the contents of `os_bootinfo` into `bootloader` without causing breakage, in case we want to do that in the future. What do you think @IsaacWoods?


Oh and one thing I wanted to note:

> We're already causing breakage

We're causing breakage by the `bootimage` update, but this `bootloader` PR should be backwards compatible. (The only change that could be breaking is that interrupts are now disabled instead of masked, but I don't think that anyone relied on this.)

IsaacWoods:
Ah that makes sense - we haven't discussed how we're going to approach multiple bootloaders in the future (we'll need 2-3 for x86 alone eventually - the current BIOS HDD image, EFI and maybe El-Torito). I'd like to see if we can actually support them all within the one crate, and just use cfg to mix the code we need into each version, which would be configurable via `bootimage`, so the dependency would always be this crate. Until that comes to fruition however, keeping `os_bootinfo` around doesn't hurt, and I support not re-exporting it :)

Overall tho, this is great work and much easier to understand what's going on, thanks Phil!

lachlansneff:
To add onto what @IsaacWoods said, we'll also need bootloaders for risc-v and the various flavors of arm eventually. It would be nice to contain them all in one crate and use cfg features or even just the compiler target to enable to correct one.

phil-opp:
@IsaacWoods 

> I'd like to see if we can actually support them all within the one crate, and just use cfg to mix the code we need into each version, which would be configurable via bootimage, so the dependency would always be this crate.

Now that the bootloader is a normal dependency, we could even do this through normal cargo features, e.g. an `bios` feature, an `efi` feature, etc.

> Overall tho, this is great work and much easier to understand what's going on, thanks Phil!

Glad to hear that, thank you!

phil-opp:
@lachlansneff Sounds good!

phil-opp:
Published as version 0.2.0

PR #21, opened 2018-07-30
Make printing from the bootloader cleaner

IsaacWoods:
I noticed that debugging from the bootloader was a bit tedious as the printer didn't allow nice formatting or handle newlines well. This PR makes the printer easier to work with and modularises it to allow printing to either the VGA buffer, the COM1 serial port, or both (configurable by features).

The configuration is mainly for future proofing (when we implement graphics mode switching, we'll want to be able to turn off the text-mode VGA printing).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,6 +14,7 @@ version = "0.2.0-beta"
 dependencies = [
  "fixedvec 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "os_bootinfo 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "spin 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "usize_conversions 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "x86_64 0.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "xmas-elf 0.6.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


phil-opp:
Awesome, thanks a lot @IsaacWoods!

I tested it locally and I get some strange behavior:

- The screen is only cleared when a println is executed. I assume that otherwise the lazy static initialization simply does not happen.
- I don't see anything when I do a println. Maybe it is optimized away because the writes are not volatile?

IsaacWoods:
> The screen is only cleared when a println is executed. I assume that otherwise the lazy static initialization simply does not happen.

Yep, the printer will not even be created unless we print. I thought that was attractive, but it does complicate stuff, since it doesn't clear the VGA buffer or initialise the serial port if we happen to not print. A simple solution would be to always print a line, even on a successful boot, but that doesn't seem like a very clean solution. I might make the initialisation explicit again, then.

> I don't see anything when I do a println. Maybe it is optimized away because the writes are not volatile?

Hmm, even compiling everything in release mode, I can't reproduce this. I won't have time to work on anything in the near future, but will have a look when I get back from holiday.

phil-opp:
>> I don't see anything when I do a println. Maybe it is optimized away because the writes are not volatile?
> 
> Hmm, even compiling everything in release mode, I can't reproduce this.

Sorry, I forgot to remove the printing from the kernel so that the printed lines were shifted off screen. So printing works as intended. However I still think that the writes should be volatile.

I did some more testing and found some other issues:

- The macro imports the `Write` trait into the parent namespace. This leads to duplicate import errors when doing multiple printlns. To fix this we could wrap the macro expansion in an additional `{}` block.
- Is there a reason that there's no `print` macro?
- When invoking multiple `println`s (after fixing the import error), the kernel does no longer boot. For example this happens on 15 printlns.
- When doing more printlns (e.g. 23), a boot loop occurs.

> I won't have time to work on anything in the near future, but will have a look when I get back from holiday.

Thanks! Have a nice vacation!

phil-opp:
The crate has changed much since this PR was created, so I'm going to close this as stale. Feel free to open a new pull request if you're still interested in these changes!

PR #22, opened 2018-08-15
Fix duplicated nll feature and removed stabilized features.

lachlansneff:


phil-opp:
Thanks!

phil-opp:
Published as version 0.2.1

PR #23, opened 2018-08-17
Integrate os_bootinfo 0.3.0 as a submodule (with some changes)

phil-opp:
- Remove VERSION constant: The kernel can no longer have a separate dependency on os_bootinfo, so it is guaranteed that the bootinfo versions match.
- Implement the From trait to convert between the frame range types of bootinfo and x86_64.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,7 +13,6 @@ name = "bootloader"
 version = "0.2.1"
 dependencies = [
  "fixedvec 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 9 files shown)


phil-opp:
@lachlansneff Does this fix nebulet?

lachlansneff:
This looks good. Until the upstream bootloader supports packaging, I'll still be using a custom fork, but this is a good step in that direction.

phil-opp:
Ok, thanks for the review!

phil-opp:
Published as 0.3.0

PR #27, opened 2018-10-23
Fix Println routine going beyond first string terminator

mihemihe:
If we do not put a RET at the end of PRINTLN, it will keep looping after printing the return carriage, executing LODSB (incrementing si and di), and printing chars until another \0 is found in the rubbish values in memory after the string indexed; most probably the end of next string written sequentially in memory or any random 0 found out there. The second \0 will actually exit PRINTLN because the caller is PRINTLN so the RET in print_done will finish PRINTLN, not like the first \0 found where the caller is still the sub-routine PRINT

beyond-string-terminator

### DIFF
--- a/src/boot.s
+++ b/src/boot.s
@@ -111,6 +111,7 @@ println:
     call print_char
     mov al, 10 # \n
     jmp print_char
+    ret
 
 # print a string
 # IN


phil-opp:
Thanks for the pull request! I'm not sure whether I understand this issue. We do a `jmp` (not a `call`) to print the `\n` in line 113, so the `ret` in `print_char` should return back to the caller of `println` (because no additional return address is pushed for `jmp`). Or am I missing anything?

Either way, adding a `ret` there shouldn't do any harm. Maybe we should also change the `jmp` to a `call` in the previous line to make the code more clear.

mihemihe:
You are right, the second one is just a jmp and then there is a ret inside..... it could be something related to my dev environment then. I will test it again during the next days...

phil-opp:
Thanks for the update!

PR #28, opened 2018-11-08
Use LLD for linking, fix page table zeroing, increase memory map size, improve errors

phil-opp:
This PR combines some updates to the bootloader:

- **More stages**: I split the second stage into three separate stages (16-bit, 32-bit, 64-bit) to make them shorter.
- **Use LLD for linking**: I was able to identify and rewrite the problematic relocations that prevented us from linking with LLD. This means that we no longer need bootloader_precompiled, as the bootloader should now compile on Linux, macOS, and Windows.
- **Go through protected mode** instead of jumping directly from real mode to long mode. This is not necessary, but I had some problems to run the bootloader on real hardware, so I tried to keep things the standard way (the problem turned out to be somewhere else).
- **Fix page table zeroing**: The code for zeroing the initial page tables didn't zero the P1 table. This was a very ugly bug, because QEMU zeroes memory by default so that it only caused errors on real hardware.
- **Increase memory map size** to 64 entries. My laptop reports more than 32 memory regions, which caused the boot to fail.

There were also some minor changes such as updating the Cargo.lock and some changes to make the bootloader more robust.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,7 +5,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bitflags"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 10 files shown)


phil-opp:
Published as version 0.3.3.

PR #29, opened 2018-11-21
Fix spurious panic on some hardware

SplittyDev:
Apparently, it is possible that non-usable memory regions overlap each other.
This is fixed simply by removing the detection of overlapping
non-usable memory and the associated panic.

Bump patch version to 0.3.5.
The change should be fully backwards-compatible.


phil-opp:
Thanks!

phil-opp:
Released as 0.3.5

PR #31, opened 2018-11-23
add Makefile

acheronfail:
This PR adds a `Makefile` that works both on Linux and macOS (so we don't have to provide instructions in the `README.md`).

Make commands:

* `make`: defaults to `make run`
* `make build`: runs `xargo build`
* `make objcopy`: runs `objcopy` (on linux and macOS)
* `make run`: runs `qemu-system-x86_64`
* `make clean`: runs `rm -f bootimage.bin`
* `make cleanall`: runs `rm -rf target` (and the `make clean` target)

Not sure if you want to use `make` to build/run this project, since showing the build commands is also informative. Let me know how what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target/
 **/*.rs.bk
+bootimage.bin
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,15 @@ language: rust
 rust:
 - nightly
 
-os: linux
+matrix:
+  include:
+    - os: linux
+    - os: osx
+
+addons:
+  homebrew:
+    packages:
+    - binutils
 
 cache:
   directories:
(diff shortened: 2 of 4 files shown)


acheronfail:
I also added a macOS build to the Travis config as well. :slightly_smiling_face:

phil-opp:
Thanks for the pull request. I'm not sure whether it's a good idea to depend on `make` for building because it is an explicit goal of this crate to be buildable with only Rust and no other dependencies. 

However, I agree that the current situation needs improvements: The Readme is outdated (e.g. we now link with LLD), the travis build only runs on Linux, the build instructions are not complete (the bootloader won't boot successfully without a kernel), the build still uses `xargo` instead of `cargo-xbuild`, and the actual build implementation is split between this crate and bootimage. I'll create a tracking issue with all the tasks that need to be done.

acheronfail:
>I'm not sure whether it's a good idea to depend on `make` for building because it is an explicit goal of this crate to be buildable with only Rust and no other dependencies.

No worries! I thought that might be the case.
I'd be happy to help out without those other issues though! 👍

phil-opp:
@acheronfail Thanks! I opened https://github.com/rust-osdev/bootloader/issues/32 to track these issues

PR #34, opened 2018-11-24
Create new `builder` sub-crate

### LINKED ISSUE #32: Improve Build System and Documentation
Things we have to do in order to improve this crate:

- [ ] Use `cargo-xbuild` instead of `xargo` (since the latter is no longer maintained)
- [ ] Run travis CI build on Linux, Windows, and macOS (only the cargo build step for now, not the objcopy step)
- [ ] Create a `build` sub-crate that builds the bootloader and links it with a kernel to create a bootable disk image. The tool needs to do the following:
  - Run `cargo xbuild --target x86_64-bootloader.json --release`
  - Convert the resulting ELF file to a binary file. Either in Rust code (like bootimage does) or through the bundled `llvm-objcopy` and [cargo-binutils](https://github.com/rust-embedded/cargo-binutils)
  - Append a "kernel info block" to the binary file: A 512 byte structure which contains the total size of the kernel (this is used by the bootloader to figure out how many bytes it needs to load into memory)
  - Append an ELF kernel right behind the kernel info block. The build tool could e.g. take the path to the ELF file as an argument.
- [ ] Update the Readme

phil-opp:
Fixes #32 

In addition to the tasks in #32, this also adds a small example kernel for testing the bootloader on travis.

cc @acheronfail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ addons:
 
 install:
   - if [ $TRAVIS_OS_NAME = osx ]; then brew update; brew install qemu; fi
-  - if [ $TRAVIS_OS_NAME = windows ]; then wget https://qemu.weilnetz.de/w64/2018/qemu-w64-setup-20180801.exe; 7z x qemu-w64-setup-20180801.exe; fi
+  - if [ $TRAVIS_OS_NAME = windows ]; then choco install qemu; export PATH="/c/Program Files/qemu:$PATH"; fi
 
 before_script:
 - rustup component add rust-src
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,11 @@ addons:
   apt:
     packages:
       - qemu-system-x86
+  homebrew:
+    packages:
+      - qemu
 
 install:
-  - if [ $TRAVIS_OS_NAME = osx ]; then brew update; brew install qemu; fi
   - if [ $TRAVIS_OS_NAME = windows ]; then choco install qemu; export PATH="/c/Program Files/qemu:$PATH"; fi
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,30 @@ language: rust
 rust:
 - nightly
 
-os: linux
+os:
+  - linux
+  - osx
+  - windows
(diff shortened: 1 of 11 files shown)


acheronfail:
**TLDR**: this will fix the Windows build errors:

```diff
 install:
  - if [ $TRAVIS_OS_NAME = osx ]; then brew update; brew install qemu; fi
- - if [ $TRAVIS_OS_NAME = windows ]; then wget https://qemu.weilnetz.de/w64/2018/qemu-w64-setup-20180801.exe; 7z x qemu-w64-setup-20180801.exe; fi
+ - if [ $TRAVIS_OS_NAME = windows ]; then choco install qemu; export PATH="/c/Program Files/qemu:$PATH"; fi
```

---

The windows build is failing because it can't find the installed `qemu-system-x86_64` executable (you probably already know but `127` means the command wasn't found). 

It works if you use `./qemu-system-x86_64` instead. 

We could also simplify the install step on windows since Travis includes Chocolately, and do something like:

```yaml
install:
  - if [ $TRAVIS_OS_NAME = windows ]; then choco install qemu; fi
```

Which installs qemu into `C:\Program Files\qemu\qemu-system-x86_64`.

If we then update the `PATH`:

```yaml
install:
  - if [ $TRAVIS_OS_NAME = windows ]; then choco install qemu; export PATH="/c/Program Files/qemu:$PATH"; fi
```

Then we can just call `qemu-system-x86_64` as normal (without the preceding `./`) and [the build should now succeed](https://travis-ci.org/acheronfail/bootloader/jobs/459308922#L60). :)

acheronfail:
For the macOS build, is there a reason you use:

```yaml
install:
  - if [ $TRAVIS_OS_NAME = osx ]; then brew update; brew install qemu; fi
```

Rather than:

```yaml
addons:
  homebrew:
    packages:
      - qemu
```

?

...IMO the latter is cleaner, since Travis takes care of it instead of us - and it also saves us an extra step in the install section.

phil-opp:
Thanks so much for your help!

phil-opp:
Published as version 0.3.6

PR #36, opened 2018-12-02
convert stage_4 of bootloader to rust

acheronfail:
I've been playing around trying to convert some assembly to Rust in this repository, and I've (almost) managed to convert `stage_4` to a Rust function.

The only thing that I can't figure out, is how to calculate the `memory_map_entry_count` argument? I've tried various things and maybe I don't understand something here. 🤔 

IIUC `memory_map_entry_count` is defined here in `stage_4.s`:

```asm
movzx rcx, word ptr mmap_ent
; If I understand correctly, the `movzx` instruction above is equivalent to:
; xor rcx, rcx
; mov  cx, word ptr mmap_ent
```

But, I've even tried changing that completely (to things like `mov rcx, 0x0`) and the result is always `6` when `load_elf` is called, so I'm really quite confused.

In any case, I'm putting this PR up to see if you can help me understand that part above, and to try and remove more assembly in favour of more Rust. 😄


src/main.rs
```diff
@@ -63,6 +62,43 @@ impl IdentityMappedAddr {
...
+
+#[no_mangle]
+pub unsafe extern "C" fn stage_4() -> ! {
+    asm!("mov bx, 0x0" :::: "intel");
+    asm!("mov ss, bx" :::: "intel"); // Set stack segment
+
+    let kernel_start = 0x400000;
+    let kernel_size = _kib_kernel_size as u64;
+    let memory_map_addr = &_memory_map as *const _ as u64;
+    let memory_map_entry_count = 6;
```
acheronfail: This is just hard-coded: from my PR description you can see I don't know how this value always resolves to `6`.

phil-opp:
Thanks for the PR! Looks good to me overall.

> But, I've even tried changing that completely (to things like mov rcx, 0x0) and the result is always 6 when load_elf is called, so I'm really quite confused.

This seems really strange.

acheronfail:
I've figured out why changing the instructions wasn't working - I was using `;` to comment out instructions, but as it turns out that doesn't actually comment them out. Using `#` instead does. So that basically means that the original code I thought was commented out was actually running. 🤦‍♂️ 

Knowing that, I can finish this PR - I'll convert the rest of `stage_4` to Rust and update this.

acheronfail:
What a silly error that was!

In any case, I think this should be good to go now. 👍

src/main.rs
```diff
@@ -63,8 +62,44 @@ impl IdentityMappedAddr {
...
+    static __page_table_end: usize;
+    static __bootloader_end: usize;
+    static __bootloader_start: usize;
+}
+
 #[no_mangle]
-pub extern "C" fn load_elf(
+pub unsafe extern "C" fn stage_4() -> ! {
+    asm!("mov bx, 0x0" :::: "intel");
+    asm!("mov ss, bx" :::: "intel"); // Set stack segment
```
phil-opp: Could you merge these two into a single `asm!` statement? This ensures that they are executed right after each other. Also, we should add `bx` to the clobber list.

phil-opp:
> I was using `;` to comment out instructions, but as it turns out that doesn't actually comment them out. Using `#` instead does. So that basically means that the original code I thought was commented out was actually running. man_facepalming

Ah yes, inline assembly is for some reasons a bit different to nasm assembly. No idea why they have different comment styles.

PR #39, opened 2018-12-06
rustify `context_switch`

strake:


phil-opp:
Thank you!

phil-opp:
Published as 0.3.9

phil-opp:
Oh, I just found an issue with this PR. The problem is the `.spin: jmp spin` part: Inline Assembly has no special support for `.` labels, so `.spin` is just a global label with the name `.spin`. But `jmp spin` jumps to the completely different `spin` label, which is defined somewhere in the assembly code. I created a fix at https://github.com/rust-osdev/bootloader/pull/40. Sorry that I didn't see it in my first review.

PR #40, opened 2018-12-08
Fix context_switch spin jump

phil-opp:
Fixes a small problem of #39: The problem is that inline Assembly has no special support for `.` labels, so `.spin` is just a global label with the name ".spin". But `jmp spin` jumps to the completely different "spin" label, which is defined somewhere in the assembly code. This PR fixes that problem by reintroducing the `context_switch_spin` label that was used in the assembly code before.


phil-opp:
Published as 0.3.10

PR #41, opened 2018-12-08
privatize spin label

strake:
Ensures no symbol name clashes.


phil-opp:
Nice! I didn't know about ` ${:private}` and `${:uid}`. Thanks!

phil-opp:
Published as 0.3.11

PR #42, opened 2019-01-18
Updated x86_64 to 0.3.5

GrayJack:
Doesn't seen to have any breakage compiling or using this crate on other projects


phil-opp:
Thanks!

Issue #44, opened 2019-02-03
Hangs before second stage on a real machine, works fine in QEMU


johanhenriksson:
Im following the OS dev tutorial series and decided to try to run the kernel on a real machine, but it hangs indefinitely before the second stage of the bootloader.

It prints ``Booting... (first stage)`` and then hangs. No other output unfortunately.

I've tried building it in release mode (as suggested in a similar issue) but it doesn't make a difference. Everything works fine when I run it in QEMU, release or debug.

Any ideas? Would be cool to see it running on real hardware :)

Edit: I'm on macOS, booting from a USB stick. Writing the image with ``dd if=target/x86_64-blog_os/release/bootimage-blog_os.bin of=/dev/disk3 && sync``

phil-opp:
I cannot reproduce this unfortunately. For me it runs fine on real hardware.

It seems like the bootloader does something that does not work on your hardware, which is difficult to debug from my end. Could you try to do some debugging to find out on which instruction it hangs? The easiest way to find this out is:

- Clone this bootloader project.
- Replace the `bootloader = 0.3.0` dependency in your kernel with a path dependency: `bootloader = { path = "../bootloader" }` (use the path to directory where you cloned the bootloader). Now the kernel should use your local bootloader for building.
- Edit the `src/stage_1.s` assembly file to find out where the error occurs
  - The `Booting... (first stage)` message is printed in [line 26](https://github.com/rust-osdev/bootloader/blob/master/src/stage_1.s#L25-L26) so we know that the problem is on a later line.
  - Try to insert the following lines at various places in the file to print the `Booting... (first stage)` message a second time:
  
  ```
    push esi    # backup `esi` because we overwrite it
    push eax    # backup `eax` `real_mode_println` overwrites it
    lea si, boot_start_str    # load the address of the message into `si`
    call real_mode_println    # print the message a second time
    pop eax    # restore `eax`
    pop esi    # restore `esi`
  ```

- Move these lines around and try to isolate the problematic instruction this way. If you see the message twice, all preceding instructions worked. If you see the message only once, the problem is in a preceding instruction.

johanhenriksson:
Sure, I will give it a shot tonight

phil-opp:
Thanks!

johanhenriksson:
[...]

PR #45, opened 2019-02-09
Only enable A20 if needed

phil-opp:
Also avoid writing bit 0.

See https://www.win.tue.nl/~aeb/linux/kbd/A20.html

Fixes an issue reported in https://github.com/rust-osdev/bootloader/issues/44


phil-opp:
Published in version 0.3.13

PR #52, opened 2019-03-31
Remove stabilized `try_from` feature

kvart2006:
This is an obsoleted feature.

warning: the feature `try_from` has been stable since 1.34.0 and no longer requires an attribute to enable
 --> src/main.rs:3:12
  |
3 | #![feature(try_from)]
  |            ^^^^^^^^
  |
  = note: #[warn(stable_features)] on by default


phil-opp:
Thanks!

bors r+

PR #53, opened 2019-04-01
Updates for new build system

phil-opp:
Update the documentation for the new build system and add an `example-kernel` that provides an easy way to try the `bootloader`.

Builds upon #51.


phil-opp:
Published as version 0.5.0

PR #55, opened 2019-04-26
Version 0.6.0

phil-opp:
The main change of this PR is to remove the `#[cfg(not(test))]` attribute from the `entry_point` macro. This makes it possible to use the macro together with custom test frameworks and `cargo xtest`. Since this is a breaking change, this PR bumps the version to 0.6.0.

This PR also adds two additional checks to the build script:

- check that the kernel exists early to provide a better error message
- check that the kernel has a non-empty text section
  - an empty text section can occur if no entry point is set
  - this turns a runtime error (bootloader panics when loading the kernel) into a compile time error

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,7 +20,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bootloader"
-version = "0.5.3"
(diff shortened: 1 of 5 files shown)


phil-opp:
Published as version 0.6.0

PR #58, opened 2019-06-21
Allow overriding the physical memory offset through an environment variable

phil-opp:
The plan is that the `bootimage` crate gets an additional `Cargo.toml` config key for this property and sets the environment variable automatically.

cc @64


64:
Looks great! Thanks.

phil-opp:
Seems spurious (chocolatey timeout)

bors retry

phil-opp:
Failed presumably because of https://status.chocolatey.org/issues/2019-06-24-chocolatey-org-not-available-from-aws/.

phil-opp:
Seems to be working again

bors retry

PR #59, opened 2019-06-23
Strip debug symbols from kernel

phil-opp:
This PR strips debug symbols from the kernel binary to speed up loading. This reduces the boot image size from 5MB to 270KB for blog_os.

(The code already works, I just want to clean it up a bit and improve the naming of the string/path variables before merging.)

cc @64

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,5 @@
 - Make the physical memory offset configurable through a `BOOTLOADER_PHYSICAL_MEMORY_OFFSET` environment variable ([#58](https://github.com/rust-osdev/bootloader/pull/58)).
+- Use a stripped copy of the kernel binary (debug info removed) to reduce load times ([#59](https://github.com/rust-osdev/bootloader/pull/59)).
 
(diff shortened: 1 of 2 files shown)


bjorn3:
Could you please add an option to disable this. Gdb attached to qemu can load debuginfo too (havent tested, but people say it works)

64:
If done properly, this shouldn't affect debugging with GDB. As I understand it, there'd be two copies of the binary - one with debug info and one without. The bootloader would simply load the one without debug info, but you could just point GDB to the one with debug info and it'd work properly. @bjorn3

bjorn3:
Thanks @64 I thought the original was stripped in place.

phil-opp:
Yeah, the original binary is not modified.

phil-opp:
Failed presumably because of https://status.chocolatey.org/issues/2019-06-24-chocolatey-org-not-available-from-aws/

phil-opp:
Published as version 0.6.1

Issue #49, opened 2019-02-21
gets stuck at "Booting first stage..." when chainloading via grub in an iso


xacrimon:
Done some testing. Seems to happen after entering protected mode. CPU at 100% and frozen

phil-opp:
Thanks for testing! I replied to you [on gitter](https://gitter.im/rust-osdev/bootloader?at=5c65d723dc3f0523ccb7722e).

phil-opp:
@xacrimon Any updates on this?

xacrimon:
Haven't had time.

xacrimon:
Ill do some more research as soon as i have time. Probably wont be until May comes around.

phil-opp:
Ok, thanks for the update!

xacrimon:
Alright, haven't had much time but it seems to be a failure related to loading the rest of the bootloader @phil-opp

xacrimon:
This is the output while running with `qemu-system-x86_64 -cdrom tulip.iso`

```
Booting (first stage)...
Failed to load rest of bootloader
```

and then it hangs

xacrimon:
Alright. Ive tried to narrow the instruction down but i have had a hard time. I know `spin` doesnt get called since i dont get more output when placing the print instructions there. But everywhere else in stage1.asm seems to work. Wierd

xacrimon:
Ah okay. If i dont supply the -cdrom flag to grub it doesnt call spin and get stuck with no error message. If i supply it it gives the load error and calls spin

xacrimon:
Alright. Done some more narrowing down. It seems to fail silently and fall into a spin loop or something in stage_1.s line 105 where it  jumps to stage 2.

```
104: lea eax, [stage_2]
105: jmp eax <- hangs here
```

xacrimon:
but as i said, it doesnt get to print the booting second stage message so something going wrong when jumping.

xacrimon:
It might be related to having issues loading. Itya load some data that isn't the bootloader since locations get messed up when packed in an iso

phil-opp:
> It might be related to having issues loading. It may load some data that isn't the bootloader since locations get messed up when packed in an iso

Yes, that could be the problem. Could you try removing the [`"relocation_model": "static"` line](https://github.com/rust-osdev/bootloader/blob/f770e97f6c52420212ff69e7acfec6b4854a46ff/x86_64-bootloader.json#L20) from the `x86_64-bootloader.json` file?

[...]

PR #60, opened 2019-06-23
VESA video mode - 800x600x16bit

jabedude:
Hello all, this PR checks if VESA 2.0 or 3.0 is supported and then enables it with fixed 800x600. This is very rough, so please let me know what else we need to do.

TODO:

- [x] Proper color handling
- [ ] Query the framebuffer address from hardware instead of hardcoding it (if possible)
- [ ] Pass the framebuffer address in BootInfo

### DIFF
--- a/src/printer/vesa_800x600.rs
+++ b/src/printer/vesa_800x600.rs
@@ -1,0 +1,81 @@
+use core::fmt::{Result, Write};
+use core::slice;
+use core::sync::atomic::{AtomicUsize, Ordering};
+
+const VGA_BUFFER: *mut u16 = (500 * 512 * 4096) as *mut _;
+const SCREEN_WIDTH: usize = 800;
+const SCREEN_HEIGHT: usize = 600;
+
+pub static X_POS: AtomicUsize = AtomicUsize::new(1); // must not be 0 so that we don't have a .bss section
+pub static Y_POS: AtomicUsize = AtomicUsize::new(1); // must not be 0 so that we don't have a .bss section
+
+pub struct Printer;
+
+impl Printer {
+    pub fn clear_screen(&mut self) {
+        let vga_buffer = Self::vga_buffer();
+        for pixel in vga_buffer {
+            *pixel = 0x66;
+        }
+        X_POS.store(0, Ordering::SeqCst);
+        Y_POS.store(0, Ordering::SeqCst);
+    }
+
+    fn vga_buffer() -> &'static mut [u16] {
+        unsafe { slice::from_raw_parts_mut(VGA_BUFFER, SCREEN_WIDTH * SCREEN_HEIGHT) }
+    }
+
+    fn newline(&mut self) {
+        let y_pos = Y_POS.fetch_add(8, Ordering::SeqCst);
+        X_POS.store(0, Ordering::SeqCst);
+        if y_pos >= SCREEN_HEIGHT {
+            self.clear_screen();
+        }
+    }
+
+    fn write_char(&mut self, c: char) {
+        use font8x8::UnicodeFonts;
+
+        if c == '\n' {
+            self.newline();
+            return;
+        }
+
+        let vga_buffer = Self::vga_buffer();
+
+        let x_pos = X_POS.fetch_add(8, Ordering::SeqCst);
+        let y_pos = Y_POS.load(Ordering::SeqCst);
+
+        match c {
+            ' '..='~' => {
+                let rendered = font8x8::BASIC_FONTS
+                    .get(c)
+                    .expect("character not found in basic font");
+                for (y, byte) in rendered.iter().enumerate() {
+                    for (x, bit) in (0..8).enumerate() {
+                        if *byte & (1 << bit) == 0 {
+                            continue;
+                        }
+                        let color = 0xff;
+                        vga_buffer[(y_pos + y) * SCREEN_WIDTH + x_pos + x] = color;
+                    }
+                }
+            }
+            _ => panic!("unprintable character"),
+        }
+
+        if x_pos + 8 >= SCREEN_WIDTH {
+            self.newline();
+        }
+    }
+}
+
+impl Write for Printer {
+    fn write_str(&mut self, s: &str) -> Result {
+        for c in s.chars() {
+            self.write_char(c);
+        }
+
+        Ok(())
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target/
 **/*.rs.bk
+*.swp
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ llvm-tools = "0.1"
 [features]
 default = []
 vga_320x200 = []
+vesa_800x600 = []
 recursive_page_table = []
 map_physical_memory = []
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -93,6 +93,7 @@ pub unsafe extern "C" fn stage_4() -> ! {
     let bootloader_start = &__bootloader_start as *const _ as u64;
     let bootloader_end = &__bootloader_end as *const _ as u64;
 
+
     load_elf(
         IdentityMappedAddr(PhysAddr::new(kernel_start)),
         kernel_size,
@@ -291,6 +292,8 @@ fn load_elf(
     }
 
     let entry_point = VirtAddr::new(entry_point);
+
+    panic!("Test");
 
     unsafe { context_switch(boot_info_addr, entry_point, stack_end) };
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -31,9 +31,11 @@ global_asm!(include_str!("stage_2.s"));
 global_asm!(include_str!("e820.s"));
 global_asm!(include_str!("stage_3.s"));
 
+#[cfg(feature = "vesa_800x600")]
+global_asm!(include_str!("video_mode/vesa_800x600.s"));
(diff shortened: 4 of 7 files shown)


phil-opp:
Thanks a lot for tackling this!

We don't want to hard code the video mode for the user, but instead make it configurable through cargo features. We currently support [vga text mode and a simple vga graphics mode (320x200)](https://github.com/rust-osdev/bootloader/blob/4f87e5a2efc239ca301fcf470781a0ff66ce268b/src/main.rs#L34-L37). Could you place the VESA code in a new file in `src/video_mode` and add a cargo feature for enabling it, like it is done for the other video modes?

jabedude:
Sure thing!

phil-opp:
Thanks, looks good now!

Were you able to print something to the screen through this? I just tried to update the `src/printer/mod.rs` module (used for printing panic messages) with a new vesa printer, but I didn't know the framebuffer location.

jabedude:
I meant to ping you, I actually was hoping to get some help with testing printing with the new video mode. I should be able to get you the framebuffer address.

phil-opp:
Sure!

Printing should work similar to the [vga_320x200](https://github.com/rust-osdev/bootloader/blob/master/src/printer/vga_320x200.rs) module, we only need the framebuffer address and the pixel format (i.e. how many bytes per pixel, how are the RGB channels encoded).

jabedude:
I found the framebuffer address for qemu: 0xFD000000. The pixel format is 5 red bits, 6 green, and 5  blue bits. I'm still having issues trying to modify vga_320x200 to print correctly

phil-opp:
@jabedude Thanks! The problem was that we need to add a page table mapping for the physical address `0xFD000000` to the page table. I pushed a commit (I hope that is fine with you) that maps the framebuffer to the 500th entry of the level 2 page table. It also adds a new printer, which is the vga_320x200 printer with some modifications. Now it looks like this:

image

I didn't add any code for color handling, so the colors are a bit strange. I deliberately didn't paint the screen back in `clear_screen` so that we see that really the whole screen is blanked.

ethindp:
Here's what I'm wondering: will this bootloader be modified sometime
to make it possible to boot on real hardware? It seems like the
bootloader goes off virtualized/emulated addresses, at least for
anything that's not 0xb8000. I would love to be able to boot my kernel
on real hardware with this thing!
If not, how can I get it to do that?

On 7/6/19, Philipp Oppermann <notifications@github.com> wrote:
> @jabedude Thanks! The problem was that we need to add a page table mapping
> for the physical address `0xFD000000` to the page table. I pushed a commit
> (I hope that is fine with you) that maps the framebuffer to the 500th entry
> of the level 2 page table. It also adds a new printer, which is the
> vga_320x200 printer with some modifications. Now it looks like this:
>
> image
>
> I didn't add any code for color handling, so the colors are a bit strange. I
> deliberately didn't paint the screen back in `clear_screen` so that we see
> that really the whole screen is blanked.
>
> --
> You are receiving this because you are subscribed to this thread.
> https://github.com/rust-osdev/bootloader/pull/60#issuecomment-508908451


-- 
Signed,
Ethin D. Probst

phil-opp:
@ethindp For me the bootloader already works on real hardware. What problems do you have exactly?

Also, could you elaborate your statement about "virtualized/emulated addresses"? Since you replied to this pull request, I assume that you mean using the address `0xFD000000`? This is just a stop-gap solution for experimentation, the plan is to query this address from the VESA later.

ethindp:
Aha, I see. Well, I was following yorublog_os tutorial and wanted to
make an ISO with it. I might've passed the wrong args to mkisofs, but
I created the IsO and the bootloader got stuck on stage 1 (I think
that there is an issue for this).

On 7/6/19, Philipp Oppermann <notifications@github.com> wrote:
> @ethindp For me the bootloader already works on real hardware. What problems
> do you have exactly?
>
> Also, could you elaborate your statement about "virtualized/emulated
> addresses"? Since you replied to this pull request, I assume that you mean
> using the address `0xFD000000`? This is just a stop-gap solution for
> experimentation, the plan is to query this address from the VESA later.
>
> --
> You are receiving this because you were mentioned.
> https://github.com/rust-osdev/bootloader/pull/60#issuecomment-508941217


-- 
Signed,
Ethin D. Probst

phil-opp:
Have you tried the instructions at https://os.phil-opp.com/minimal-rust-kernel/#real-machine? The `bootimage` tool already creates a hard disk image, so you don't need `mkisofs` for this (I'm not sure about CD images though).

As a side note: Can we continue the conversation in a new issue? This issue issue is about adding support for the VESA video mode, so I think a separate issue would fit better.

jabedude:
@phil-opp That is awesome! Of course, I was hoping to collaborate in this PR (I have no prior experience with VESA/BIOS printing).

phil-opp:
@jabedude Great! I updated the PR description with a todo list. Do you think anything else is missing?

jabedude:
The TODOs look good to me! I pushed a commit to query the framebuffer address, but I'm not sure the best way to test that.

Related to the third TODO, how can I get access to the `VESAfb` variable from `load_elf()`?

phil-opp:
I think we can do the same as in https://github.com/rust-osdev/bootloader/blob/b5d43caa72173b502dd1eabcee29d534ef24741a/src/main.rs#L66-L77 to get access to `VESAfb` and then pass as argument to `load_elf`. To use it in the printer, we need to initialize it at runtime with the given framebuffer address, e.g. through an `init` method.

src/printer/vesa_800x600.rs
```diff
@@ -0,0 +1,81 @@
...
+                    for (x, bit) in (0..8).enumerate() {
+                        if *byte & (1 << bit) == 0 {
+                            continue;
+                        }
+                        let color = 0xff;
+                        vga_buffer[(y_pos + y) * SCREEN_WIDTH + x_pos + x] = color;
+                    }
+                }
+            }
+            _ => panic!("unprintable character"),
```
bjorn3: Please dont panic in this case. Maybe print `?` or something like that.

adeadman: The panic won't work as the writer is already locked for the write. One way around this deadlock situation is to separate the screen drawing from the character lookup in the font map - or to validate that all characters in the string are in the character set supported before calling into the static object.

adeadman:
> I think we can do the same as in
> 
> https://github.com/rust-osdev/bootloader/blob/b5d43caa72173b502dd1eabcee29d534ef24741a/src/main.rs#L66-L77
> 
> to get access to `VESAfb` and then pass as argument to `load_elf`. To use it in the printer, we need to initialize it at runtime with the given framebuffer address, e.g. through an `init` method.

Thinking about this, for flexibility we probably want to load an entire VBEModeInfo struct into the kernel space so we have flexibility for choosing different modes in the bootloader. 

The framebuffer mapping to virtual memory currently hardcodes the framebuffer address for qemu and the size for an 800x600x16 mode, so getting it to work with an arbitrary framebuffer address (e.g. from real hardware) using the vbe mode info is a bit harder. I've been working on it on and off for the last couple weeks and am wondering if it would be easier or even possible to do the memory mapping from the rust code, or if it has to be done while in protected mode stage 3 before we jump to the kernel loading.

ethindp:
Hate to go OT, but would it be possible for me to boot my kernel using
Grub? Or would that not work (i.e. if I were to use grub-mkrescue)?

On 8/10/19, Aaron Deadman <notifications@github.com> wrote:
>> I think we can do the same as in
>>
>> https://github.com/rust-osdev/bootloader/blob/b5d43caa72173b502dd1eabcee29d534ef24741a/src/main.rs#L66-L77
>>
>> to get access to `VESAfb` and then pass as argument to `load_elf`. To use
>> it in the printer, we need to initialize it at runtime with the given
>> framebuffer address, e.g. through an `init` method.
>
> Thinking about this, for flexibility we probably want to load an entire
> VBEModeInfo struct into the kernel space so we have flexibility for choosing
> different modes in the bootloader.
>
> The framebuffer mapping to virtual memory currently hardcodes the
> framebuffer address for qemu and the size for an 800x600x16 mode, so getting
> it to work with an arbitrary framebuffer address (e.g. from real hardware)
> using the vbe mode info is a bit harder. I've been working on it on and off
> for the last couple weeks and am wondering if it would be easier or even
> possible to do the memory mapping from the rust code, or if it has to be
> done while in protected mode stage 3 before we jump to the kernel loading.
>
> --
> You are receiving this because you were mentioned.
> https://github.com/rust-osdev/bootloader/pull/60#issuecomment-520190457


-- 
Signed,
Ethin D. Probst

phil-opp:
@ethindp See <https://github.com/rust-osdev/bootloader/blob/master/doc/chainloading.md> and https://github.com/rust-osdev/bootloader/issues/49. In the future, please just open a new issue instead of commenting on an unrelated issue.

phil-opp:
@adeadman 

>  I've been working on it on and off for the last couple weeks and am wondering if it would be easier or even possible to do the memory mapping from the rust code, or if it has to be done while in protected mode stage 3 before we jump to the kernel loading.

If we don't access it before the mapping, we can probably map it later. However, we need to adjust our panic printer for that (i.e. it shouldn't try to print to the VESA framebuffer before it is mapped).

phil-opp:
Given that the bootloader was changed significantly since this PR (e.g. in #130), I'm going to give this a close. We still don't support choosing custom framebuffer resolutions though, so if anyone wants to tackle this, please let me know!

Thanks you @jabedude for all your work!

PR #64, opened 2019-07-09
Build fixedvec from github

64:
The version of `fixedvec` on GitHub has no dependencies at all. The version on crates.io it depends on `skeptic` which pulled in a number of unnecessary libraries like `pulldown-cmark`, so we save about 30s by just building directly.

Before:
```
( cd example-kernel; cargo xbuild; cd ..; KERNEL= cargo xbuild --release; )  234.87s user 6.11s system 195% cpu 2:03.31 total
```

After:
```
( cd example-kernel; cargo xbuild; cd ..; KERNEL= cargo xbuild --release; )  197.42s user 5.01s system 175% cpu 1:55.09 total
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,7 +22,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "bootloader"
 version = "0.6.3"
 dependencies = [
- "fixedvec 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


phil-opp:
The problem is that crates with git dependencies can't be pushed to crates.io.

64:
Ah! My mistake. I'll see if I can ping one of the library authors to publish the latest version to `crates.io`.

phil-opp:
I just opened https://github.com/rust-embedded/fixedvec-rs/issues/15

phil-opp:
But I'm not sure whether someone still maintains it, so perhaps pinging some rust-embedded folks is a good idea.

PR #67, opened 2019-07-15
Use volatile accesses in VGA code and make font dependency optional

64:
1. Make VGA code use volatile accesses (in reality, won't make much of a difference, but is good practice and sets a good example for people who are reading the code).

2. Make the font dependency optional and only pull it in if the `vga_320x200` feature is enabled.


phil-opp:
Published as version 0.6.4

PR #68, opened 2019-07-18
Only include dependencies when `binary` feature is enabled

phil-opp:
This change vastly improves compile times when only `lib.rs` is compiled because no dependencies need to be compiled in this case. See https://github.com/rust-lang/cargo/issues/1982#issuecomment-319661478 for more information about the approach.

This is a **breaking change** since a build of the executable without the `binary` feature will result in an error.

Depends on https://github.com/rust-osdev/bootimage/pull/43, so we will need to increase the minimally supported bootimage version. To avoid public breakage, we should probably wait a bit after releasing the new bootimage version to give people time to upgrade.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,10 @@ repository = "https://github.com/rust-osdev/bootloader"
 edition = "2018"
 build = "build.rs"
 
+[[bin]]
+name = "bootloader"
+required-features = ["binary"]
+
 [dependencies]
 xmas-elf = { version = "0.6.2", optional = true }
 x86_64 = { version = "0.7.2", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,11 +25,11 @@ features = ["unicode"]
 optional = true
 
 [build-dependencies]
-llvm-tools = "0.1"
+llvm-tools = { version = "0.1", optional = true }
 
 [features]
 default = []
-binary = ["xmas-elf", "x86_64", "usize_conversions", "fixedvec"]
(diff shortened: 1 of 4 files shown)


phil-opp:
To try this change today:

- Install the bootimage version from https://github.com/rust-osdev/bootimage/pull/43:
  
  ```
  cargo install bootimage --git https://github.com/rust-osdev/bootimage.git --branch binary-feature --force
  ```

- Update the bootloader dependency in your Cargo.toml:
  
  ```toml
  [dependencies]
  bootloader = { git = "https://github.com/rust-osdev/bootloader.git", branch="binary-feature", features = ["map_physical_memory"]}
  ```

Now building the bootloader library should no longer require compiling the bootloader dependencies:

```
>  cargo xbuild
   […sysroot build…]

   Compiling llvm-tools v0.1.1
   Compiling bootloader v0.6.4 (https://github.com/rust-osdev/bootloader.git?branch=binary-feature#aa2bbc14)
   Compiling blog_os v0.1.0 (/home/philipp/Documents/blog_os/code)
    Finished dev [unoptimized + debuginfo] target(s) in 1.36s
```

64:
Actually there is one more thing - do we need the build script when the binary feature is disabled? We are still pulling in `llvm-tools` twice

phil-opp:
Good point! I disabled the build script and the llvm-tools dependency too now

phil-opp:
I think it's fine to merge this now. We will need to release a new semver-breaking version anyway, so no breakage should occur unless you manually increase the bootloader version in your Cargo.toml.

phil-opp:
Published as version 0.7.0

PR #69, opened 2019-07-30
Bump fixedvec to 0.2.4

64:
Improves compile times by ~15s on my machine.


phil-opp:
I already updated the Cargo.lock in https://github.com/rust-osdev/bootloader/commit/18ecb2d409a07c8d11eb5bfea1c68a9834c6e9dd, which included fixedvec 0.2.4. So I think that version 0.2.4 is already used when building the latest version of the bootloader.

64:
Ah, right.

PR #71, opened 2019-08-01
Dynamically map kernel stack, boot info, physical memory and recursive table

### LINKED ISSUE #61: Make the kernel stack movable
Right now the kernel stack is always fixed in memory. If I want to make my kernel higher-half the kernel will need to move. This can be done by providing a custom linker script to the kernel target. This works fine. The stack that is setup by the bootloader however is always fixed in memory and will not move with the kernel.

I suggest making the stack location configurable like the physical memory offset already is. A more elegant solution would be to switch to an offset and automatically calculate the stack location.

64:
Fixes #61. This provides a way for the user to set the kernel stack address through an environment variable, or calculate it dynamically, making sure that it does not interfere with any other mappings. We also calculate space for the recursive page table, boot info and physical memory offset, with the small optimisation that the boot info and kernel stack are placed adjacently if possible to save a few extra page table allocations.

Going to punt on a way to unmap these things for now. Reclaiming memory from unused page tables is a bit tricky and it would be nice to do it right.

### DIFF
--- a/src/level4_entries.rs
+++ b/src/level4_entries.rs
@@ -21,7 +21,7 @@ impl UsedLevel4Entries {
             let end_page: Page =
                 Page::containing_address(VirtAddr::new(segment.virtual_addr + segment.mem_size));
 
-            for p4_index in u64::from(start_page.p4_index())..u64::from(end_page.p4_index()) {
+            for p4_index in u64::from(start_page.p4_index())..=u64::from(end_page.p4_index()) {
                 used.entry_state[p4_index as usize] = true;
             }
         }
--- a/src/level4_entries.rs
+++ b/src/level4_entries.rs
@@ -1,0 +1,43 @@
+use core::convert::TryInto;
+use fixedvec::FixedVec;
+use x86_64::ux;
+use x86_64::{structures::paging::Page, VirtAddr};
+use xmas_elf::program::ProgramHeader64;
+
+pub struct UsedLevel4Entries {
+    entry_state: [bool; 512], // whether an entry is in use by the kernel
+}
+
+impl UsedLevel4Entries {
+    pub fn new(segments: &FixedVec<ProgramHeader64>) -> Self {
+        let mut used = UsedLevel4Entries {
+            entry_state: [false; 512],
+        };
+
+        used.entry_state[0] = true; // TODO: Can we do this dynamically?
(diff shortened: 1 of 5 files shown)


src/page_table.rs
```diff
@@ -18,16 +20,15 @@ pub(crate) fn map_kernel(
...
+    // Create a stack
     let stack_size: u64 = 512; // in pages
     let stack_end = stack_start + stack_size;
 
     let flags = PageTableFlags::PRESENT | PageTableFlags::WRITABLE;
     let region_type = MemoryRegionType::KernelStack;
 
-    for page in Page::range(stack_start, stack_end) {
+    // Leave the first page unmapped as a 'guard page'
+    for page in Page::range(stack_start + 1, stack_end) {
```
phil-opp: Nit: This makes the actual stack size one page smaller than `stack_size`. I don't think that it matters much, but how about doing a `let stack_start = stack_start + 1` before adding the `stack_size`?

src/level4_entries.rs
```diff
@@ -0,0 +1,43 @@
...
+        };
+
+        used.entry_state[0] = true; // TODO: Can we do this dynamically?
+
+        for segment in segments {
+            let start_page: Page = Page::containing_address(VirtAddr::new(segment.virtual_addr));
+            let end_page: Page =
+                Page::containing_address(VirtAddr::new(segment.virtual_addr + segment.mem_size));
+
+            for p4_index in u64::from(start_page.p4_index())..u64::from(end_page.p4_index()) {
```
phil-opp: This should be an inclusive `..=` range, otherwise we don't count any sections that live within a single P4 index.

src/page_table.rs
64: I had it like that so that it wouldn’t spill over into a new p1 table in the case that the boot info is mapped to the page right before the stack.

Unrelated: In general 512 pages for the stack seems like quite a lot; for reference the Linux kernel only uses 2 pages per kernel stack.

phil-opp: I'm happy to change that number. Ideally, this would be customizable through an environment variable too, so that we can default to a small stack and let people increase the size if needed.

64: I'll add that to the roadmap

Issue #70, opened 2019-07-31
32bit support


bjorn3:
cc https://github.com/phil-opp/blog_os/issues/403#issuecomment-516832927.

@Mendess2526 let's continue the discussion here.

This will need src/stage_3.s to be adapted to stay in protected mode and to setup non-PAE page tables. It will then need at least changes to the rust part to handle non-PAE page tables.

ethindp:
I hope this is togglable with a feature flag. Just saying, I'm
personally not in favor of this idea.

On 7/31/19, bjorn3 <notifications@github.com> wrote:
> cc https://github.com/phil-opp/blog_os/issues/403#issuecomment-516832927.
>
> @Mendess2526 let's continue the discussion here.
>
> This will need src/stage_3.s to be adapted to stay in protected mode and to
> setup non-PAE page tables. It will then need at least changes to the rust
> part to handle non-PAE page tables.
>
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/rust-osdev/bootloader/issues/70


-- 
Signed,
Ethin D. Probst

64:
Yeah it would need to be a toggleable feature. We should support the multiboot2 protocol for this

bjorn3:
> I hope this is togglable with a feature flag.

Shouldnt it just work when it is compiled for `i686` instead of `x86_64`?

64:
Yeah that’s a good point. I think in the mean time it might be good to add some check in the build script that it’s being built for 64-bit mode, otherwise error out and point to this issue

phil-opp:
>  I think in the mean time it might be good to add some check in the build script that it’s being built for x86_64 (if that is even possible), otherwise error out and point to this issue

[...]

PR #72, opened 2019-08-02
Configurable kernel stack size, better non-x86_64 errors

64:
Allows the configuration of the kernel stack size through an environment variable.

Also makes the errors a bit nicer if you try to use this crate from a non-x86_64 architecture (only works if you build it as a library), as discussed in #70.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,16 @@
 #[cfg(not(feature = "binary"))]
-fn main() {}
+#[allow(unreachable_code)]
+fn main() {
+    #[cfg(target_arch = "x86")]
+    panic!("This crate currently does not support 32-bit protected mode. \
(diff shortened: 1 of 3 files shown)


phil-opp:
Looks good to me! Nice cleanup in the build.rs.

CI is currently broken because of https://github.com/rust-lang/rust/issues/63195. Since this is a non-critical change I would probably wait a day or two until the issue is fixed instead of force-merging.

phil-opp:
I think the rust nightly should be fixed now.

bors r+

phil-opp:
@64 Could you cargo fmt this?

64:
Damn, I really need to set up a pre-commit hook or something...

PR #81, opened 2019-10-07
Change the way the kernel entry point is called to honor alignement ABI

### LINKED ISSUE #80: Incorrect stack alignment
The boot loader provides the kernel with a miss aligned stack.

This cause issues when SSE is enabled, as some of those instructions expect 16-aligned stacks and the alignment is off by eight.

Reproduction steps:

Take blog-os post-03, enable SSE in the boot loader and target, call panic! in main, if running under bochs you will get a fault.
(QEMU appears to ignore the issue)

The fault occurs on  movaps instruction form xmm0 to the (mis-aligned stack).

The stack alignment is mess-ed up because when the _start function is called the compiler assumes a return address has been pushed on the stack, making it 16x +8 aligned, while it is apparently 16 aligned when called by the bootloader.

From reading OS dev wiki, the kernel entry point should be reached using a call instruction so that it matches the compiler assumptions, (Even in C) whereas the boot loader appears to use a jmp.

GuillaumeDIDIER:
Compilers expect functions to be called with the stack containing the return address making the rsp be 8 mod 16. jmp caused rsp to be 0 mod 16 (aka 16 bytes aligned), which then caused compiler to meke the stack misaligned everywhere.

Fix #80

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -44,7 +44,7 @@ global_asm!(include_str!("video_mode/vga_320x200.s"));
 global_asm!(include_str!("video_mode/vga_text_80x25.s"));
 
 unsafe fn context_switch(boot_info: VirtAddr, entry_point: VirtAddr, stack_pointer: VirtAddr) -> ! {
-    asm!("jmp $1; ${:private}.spin.${:uid}: jmp ${:private}.spin.${:uid}" ::
+    asm!("call $1; ${:private}.spin.${:uid}: jmp ${:private}.spin.${:uid}" ::
          "{rsp}"(stack_pointer), "r"(entry_point), "{rdi}"(boot_info) :: "intel");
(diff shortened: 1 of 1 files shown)


GuillaumeDIDIER:
Apparently the test just blew up because rustfmt is unavailable in todays nightly 😱

GuillaumeDIDIER:
Note all the test passed apart from the missing rustfmt.

phil-opp:
@GuillaumeDIDIER Thanks for reporting and directly fixing this issue!

Don't worry about the failing rustfmt checks. This is a known problem and I already have a solution in mind. I'm planning to rewrite the CI scripts and update the testing framework soon, and I will fix it then.

phil-opp:
Published as version 0.8.2

PR #82, opened 2019-10-08
Add support for Github Actions

phil-opp:
Also fixes the problem mentioned in https://github.com/rust-osdev/bootloader/pull/81#issuecomment-538994575 (cc @GuillaumeDIDIER).


phil-opp:
bors try

PR #86, opened 2019-11-18
Fix CI formatting check: Use actions-rs/toolchain action for finding correct nightly

phil-opp:


phil-opp:
Merging without approving review to fix CI.

PR #91, opened 2020-01-08
Move architecture checks from build script into lib.rs

### LINKED ISSUE #90: `error: This crate currently does not support 32-bit protected mode` when compiling kernel with x86_64-unknown-none target
After updating `bootloader`  to 0.8.3 I started getting `error: This crate currently does not support 32-bit protected mode` error but I am pretty sure that I cross-compile x86_64 arch binary.
This issue is not reproducible with bootloader version 0.7.1 or lower. Here is the content of my target x86_64-unknown-none.json:
{
  "llvm-target": "x86_64-unknown-none",
  "data-layout": "e-m:e-i64:64-f80:128-n8:16:32:64-S128",
  "arch": "x86_64",
  "target-endian": "little",
  "target-pointer-width": "64",
  "target-c-int-width": "32",
  "os": "none",
  "executables": true,
  "linker-flavor": "ld.lld",
  "linker": "rust-lld",
  "panic-strategy": "abort",
  "disable-redzone": true,
  "features": "-mmx,-sse,+soft-float"
}

I am compiling with this command:
cargo bootimage --target x86_64-unknown-none.json

Best regards!

phil-opp:
The build script always gets compiled for the host platform, so it only checked that the _host_ was `x86_64`. With the check in lib.rs, we correctly verify that the target is `x86_64`.

Fixes #90

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,14 +1,5 @@
 #[cfg(not(feature = "binary"))]
-fn main() {
-    #[cfg(target_arch = "x86")]
-    compile_error!(
-        "This crate currently does not support 32-bit protected mode. \
-         See https://github.com/rust-osdev/bootloader/issues/70 for more information."
-    );
-
-    #[cfg(not(any(target_arch = "x86_64", target_arch = "x86")))]
-    compile_error!("This crate only supports the x86_64 architecture.");
-}
+fn main() {}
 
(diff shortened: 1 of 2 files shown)


phil-opp:
Thanks for reviewing!

phil-opp:
Published as version 0.8.4

PR #92, opened 2020-01-14
Update x86_64 dependency

vinaychandra:
- Removed `ux::u8`
- Switched to `UnusedPhysFrame`


phil-opp:
Published as version 0.8.5

PR #96, opened 2020-02-19
Add basic support for ELF thread local storage segments

### LINKED ISSUE #95: Enable tls sections
`tdata` and `tbss` section stores thread local storage for the system. In kernel, we will need to initialize this section per-cpu or per-thread depending on usages. Rust provides a `#[thread_local]` attribute which puts the data in the above sections.

Please enable that section in bootloader and pass its location to the kernel so that it can map the TLS data as needed.

phil-opp:
This reports the thread local storage segment of the kernel (if present) in the boot information. This should make it possible to use the unstable [`#[thread_local]` attribute](https://github.com/rust-lang/rust/issues/29594). Since the TLS template overlaps with parts of a LOAD segment, no additional mapping needs to be done.

Note that the TLS template should not be used directly. Instead, the kernel should create a copy the TLS template for each thread and then set up the `fs` register accordingly. See https://uclibc.org/docs/tls.pdf for more information.

This code only works for a single TLS segment. If more than one TLS segment are present in the executable, the bootloader errors. We can of course lift this restriction if this becomes a problem in practice.

Fixes #95

### DIFF
--- a/src/bootinfo/mod.rs
+++ b/src/bootinfo/mod.rs
@@ -42,6 +42,7 @@ pub struct BootInfo {
     /// can be safely accessed.
     #[cfg(feature = "map_physical_memory")]
     pub physical_memory_offset: u64,
+    tls_template: TlsTemplate,
     _non_exhaustive: u8, // `()` is not FFI safe
 }
 
@@ -51,11 +52,18 @@ impl BootInfo {
     #[doc(hidden)]
     pub fn new(
         memory_map: MemoryMap,
+        tls_template: Option<TlsTemplate>,
         recursive_page_table_addr: u64,
         physical_memory_offset: u64,
     ) -> Self {
+        let tls_template = tls_template.unwrap_or(TlsTemplate {
+            start_addr: 0,
+            file_size: 0,
+            mem_size: 0,
+        });
         BootInfo {
             memory_map,
+            tls_template,
             #[cfg(feature = "recursive_page_table")]
             recursive_page_table_addr,
             #[cfg(feature = "map_physical_memory")]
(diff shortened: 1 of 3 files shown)


phil-opp:
@vinaychandra Could you try whether this change works for you?

phil-opp:
Ah, it seems like I understood something wrong. I will create a new version.

phil-opp:
With the new version, I can now successfully read the TLS template info for a small example:

```rust
#[thread_local]
pub static mut DATA: u64 = 0;
#[thread_local]
pub static BSS: u64 = 0;

fn kernel_main(boot_info: &'static BootInfo) -> ! {
    println!("Hello World{}", "!");
    println!("TLS INFO: {:x?}", boot_info.tls_template());

    // this panics because we did not set up the fs register yet
    println!("DATA: {}, BSS: {}", unsafe { DATA }, BSS);
}
```

The result is the following:

image

I get the same values with `readelf -l target/x86_64-blog_os/debug/blog_os`:

```
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000200040 0x0000000000200040
                 0x00000000000001c0 0x00000000000001c0  R      0x8
  LOAD           0x0000000000000000 0x0000000000200000 0x0000000000200000
                 0x0000000000003544 0x0000000000003544  R      0x1000
  LOAD           0x0000000000004000 0x0000000000204000 0x0000000000204000
                 0x000000000001f23c 0x000000000001f23c  R E    0x1000
  LOAD           0x0000000000024000 0x0000000000224000 0x0000000000224000
                 0x00000000000012d0 0x00000000000012d0  RW     0x1000
  LOAD           0x0000000000026000 0x0000000000226000 0x0000000000226000
                 0x000000000000000e 0x0000000000002178  RW     0x1000
  TLS            0x0000000000024000 0x0000000000224000 0x0000000000224000
                 0x0000000000000008 0x0000000000000010  R      0x8
  GNU_RELRO      0x0000000000024000 0x0000000000224000 0x0000000000224000
                 0x00000000000012d0 0x0000000000002000  R      0x1
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x0

 Section to Segment mapping:
  Segment Sections...
   00     
   01     .rodata .eh_frame 
   02     .text 
   03     .tdata .data.rel.ro .got 
   04     .data .bss 
   05     .tdata .tbss 
   06     .tdata .data.rel.ro .got 
   07     
```

phil-opp:
Published as version 0.8.0

Issue #74, opened 2019-08-14
Enable support for SMP Trampoline


vinaychandra:
The current entry point is hardcoded to `_start` (or the `entry_point!` macro). This provides a single entry point to the code.

When bootstrapping SMP, the new processor starts up at a given physical address in the IPI (Inter-Processor Interrupt) in 16 bit mode. This requires us to start at 16 bit code and enable 64 bit mode again. Please provide a way to define multiple "entry points" which defaults to the current `_start` but can also be used to bootstrap SMP.

64:
I’m getting near SMP in my own kernel, so I’ll probably investigate this a bit. It might be difficult to do in a platform agnostic way. For example, how should the page tables be set up?

ethindp:
I've got PCI going and am just confused on how I should manage some
kind of driver registration framework to pass the PCI devices off to.
Heh, I don't even have networking, but SMP and APIC would be nice.

On 8/14/19, Matt Taylor <notifications@github.com> wrote:
> I’m getting near SMP in my own kernel, so I’ll probably investigate this a
> bit. It might be difficult to do in a platform agnostic way. For example,
> how should the page tables be set up?

>

> --

> You are receiving this because you are subscribed to this thread.

> Reply to this email directly or view it on GitHub:

> https://github.com/rust-osdev/bootloader/issues/74#issuecomment-521311551


-- 
Signed,
Ethin D. Probst

vinaychandra:
@64 the way I understand it, we don't need to setup new page tables, etc. We can leave other processors potentially in the same place as the bootstrap processor. The pagetables are all setup for it anyway. So, we assume the same and start up the new processor into the function provided by the user. Any other changes will have to be done by the developer.

64:
You don’t need to set up page tables, but you need to load *something* into cr3 when switching to long mode.

vinaychandra:
Agreed, that's where i said we can use the same assumptions as the bootstrap processor. We can load with the same CR3 but point EIP to the new entry point rather than the old one.

64:
Booting up the APs is usually done fairly late in the boot process. Some people may have switched to a different p4 table in this time, so I think it would be better to have some way of setting the value at runtime. But that shouldn’t be too difficult.

ethindp:
Is there a way we can make a separate rust crate for SMP and allow the
kernel to do all the SMP handling, or is that only doable by the
bootloader/BIOS/UEFI? I don't fully understand SMP myself, but having
[...]

PR #97, opened 2020-02-19
Add boilerplate for a page aligned SMP trampoline function

phil-opp:
This code crates a page aligned `smp_trampoline` function that lives in the first megabyte and exports its address in the boot information. According to https://github.com/rust-osdev/bootloader/issues/74, it should be possible to implement a SMP trampoline function using these building blocks.

### DIFF
--- a/linker.ld
+++ b/linker.ld
@@ -25,13 +25,17 @@ SECTIONS {
     . = 0x7c00;
     _stack_end = .;
 
-    .bootloader :
+    .bootloader_first_stage :
     {
         /* first stage */
         *(.boot-first-stage)
+    }
 
+    .bootloader :
+    {
         /* rest of bootloader */
         _rest_of_bootloader_start_addr = .;
+        *(.smp_trampoline)
         *(.boot)
         *(.context_switch)
         *(.text .text.*)
--- a/src/bootinfo/mod.rs
+++ b/src/bootinfo/mod.rs
@@ -29,6 +29,7 @@ pub struct BootInfo {
     /// the memory map before passing it to the kernel. Regions marked as usable can be freely
     /// used by the kernel.
     pub memory_map: MemoryMap,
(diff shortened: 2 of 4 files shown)


phil-opp:
I only did a quick test, but it appears to work:

image

linker.ld
```diff
@@ -25,13 +25,17 @@ SECTIONS {
...
     {
         /* first stage */
         *(.boot-first-stage)
+    }
 
+    .bootloader :
+    {
         /* rest of bootloader */
         _rest_of_bootloader_start_addr = .;
+        *(.smp_trampoline)
```
vinaychandra: Quick question. How is it guaranteed that this will be in the first MB? If the rest of bootloader is large, wouldn't you dynamically search for a location and use that to store all the kernel rather than placing it in the first MB?

phil-opp: We put that section together with the bootloader sections. The bootloader always lives in the first MB because otherwise it would not be able to load itself in real mode. So everthing uintl `__bootloader_end` needs to be in the first MB.

phil-opp:
I don't think that we should merge this PR without implementing the trampoline function. Feel free to open a new PR with the commit from this PR included.

vinaychandra:
I will try do that. My main idea is to make a near copy of the current stage 1 and stage 2 without much of stage 3 so that we can go into 64bit mode and enter a function as specified by the user. I will probably need help in writing a macro so that a function can be detected in user's kernel so that we can jump to that from bootloader.

phil-opp:
>  I will probably need help in writing a macro so that a function can be detected in user's kernel so that we can jump to that from bootloader.

We could use the Cargo.toml configuration to let the kernel specify the function name that the trampoline should call. This way, we could even make it a compile time error when the kernel does not set a function name.

phil-opp:
Closing, as this is only an old draft.

Issue #100, opened 2020-03-06
Allow specifying address of boot_info structure.


Darksecond:
When making a higher half kernel it would be useful to move the boot_info into the higher half. It would also have to remap all the child structures, like memory_map

I imagine it would go into the `[package.metadata.bootloader]` like `physical-memory-offset` and `kernel-stack-address`.

phil-opp:
Implemented in #101.

PR #101, opened 2020-03-07
Implement boot-info-address

Darksecond:
Resolve issue #100. This implements a `boot-info-address`, which instead of dynamically relocating the boot info struct, will place it at the provided address. This functions the same as the `kernel-stack-address` and `physical-memory-offset`.


phil-opp:
Published as version 0.8.9

Issue #99, opened 2020-03-04
VGA video modes


RKennedy9064:
I've finally had some spare time to start working on my OS again and was hoping I might have a chance to give back after all the great work you've done so far. I know there's currently an open issue for VESA video modes https://github.com/rust-osdev/bootloader/issues/37, but I didn't want to hijack that thread.

Since you have plans for VESA video modes in the future, would you also be interested in adding support for VGA modes? I've just finished a minimal example that has support for switching to `80x25 text` mode from the kernel using VGA registers. The current code can be found here https://gist.github.com/RKennedy9064/374ed861f743079e43b1c5da87af63f8.

I was able to successfully boot my OS using the `vga_320x200` mode from `bootloader`, then switch to `80x25 text` mode and use the `println!` macro to display messages again. 

I'm not sure if this is on the road map, or even something you're interested in, but if you are I'd love to possible collaborate to include something like this. My plan is to clean up the existing code, add more comments and tests and then try to add more support for additional modes. I figured it would be nice to have the ability to switch video modes from the kernel at run time. 

Let me know what you think when you have the time and thanks for all the great work so far on the tutorials!

phil-opp:
Sounds awesome! I didn't have time to explore VGA/VESA video modes recently, but I think it would be great to have support for them. Being able to switch the video mode from the kernel at run time sounds like an even better approach. So yes, I would love to merge a pull request for this!

If I remember correctly, one reason for exploring additonal video modes was that the VGA text buffer is apparently not supported with UEFI. Since we want to add UEFI support, it would be great to have the same video modes work with both BIOS and UEFI booting. Maybe you know something about this?

RKennedy9064:
I don't really have much experience with UEFI, but from what I've read it has it's own way of outputting using `OutputString`, then once you initialize everything and call `ExitBootServices` you should have full control over the system, with the ability to switch modes by programming the VGA registers. Can anyone with more knowledge of UEFI confirm or deny this?

RKennedy9064:
[...]

PR #102, opened 2020-03-11
Vga mode

RKennedy9064:
Here's the initial PR as mentioned in #99. I've commented the public facing methods so far, but wanted to wait for the private methods until this initial PR is reviewed/cleaned up since it's so big. My first goal was to get everything working, so I'm sure there are enhancements that can be made and things that can be done better.

I also had to add 2 dependencies that I used to help implement everything. 1 is `spinning_top` and the other is `conquer_once`. I'm okay with using different methods if you don't want to add those as dependencies to `bootloader`. I'm also not sure if I added them correctly to `cargo.toml`, I just followed what was already there.

One last thing to note is that addresses `0xa0000 -> 0xbffff` will need to be mapped in order for these features to work correctly. I manually mapped them in my kernel for testing, but I wasn't sure where you wanted it mapped here so I left that part out for now. Let me know what you think when you get a chance.


phil-opp:
Thanks a lot for this PR!

I only took a quick look, but it seems like this does not really need to be part of the bootloader itself since it interacts with the VGA hardware through normal port I/O (in contrast to requiring BIOS calls). So I would propose that we create a separate crate for this. If you like, we could create it under the `rust-osdev` organization, with you getting admin privileges on this repo of course.

A separate crate has the advantage that it keeps the bootloader small, so that people that don't use the VGA functionality do not need to compile it into their projects, resulting in shorter compile times. It also allows people that use other bootloaders to still use the VGA crate, so that your code is useful to more people.

What do you think?

RKennedy9064:
I'm okay with creating a separate crate. I wasn't 100% bootloader was the right spot either. Just let me know what I need to do to help get it set up. Would it make sense to at least add the address mappings for `0xa0000 -> 0xbffff` like you did for `0xb8000` to bootloader?

phil-opp:
I sent you an invite to the rust-osdev organization. After accepting, you should be able to create a new repository under the organziation.

>  Would it make sense to at least add the address mappings for `0xa0000 -> 0xbffff` like you did for `0xb8000` to bootloader?

I'm not sure whether it's a good idea to add these mappings by default because not everyone will want them. But we could add an optional feature or a configuration option for this.

phil-opp:
Let me know if you need any help setting up the repository!

RKennedy9064:
Thanks for the access. I got the repo setup and am working on configuring a workflow for automatic testing. I'd be fine with a configuration option for those mappings. It would be nice to be able to turn them on so I don't have to map it all in my tests. Is that something I can add and open a PR for? If so, where do you currently handle mapping `0xb8000` for the `80x25` frame buffer?

phil-opp:
> Thanks for the access. I got the repo setup and am working on configuring a workflow for automatic testing.

Great!



>  I'd be fine with a configuration option for those mappings. It would be nice to be able to turn them on so I don't have to map it all in my tests. Is that something I can add and open a PR for? If so, where do you currently handle mapping `0xb8000` for the `80x25` frame buffer?

Currently as part of the assembly code for the video modes: <https://github.com/rust-osdev/bootloader/search?q=vga_map_frame_buffer&type=>. The reason for this is that the bootloader itself already prints to the VGA buffer in some cases.

Looking at this code, I think it might make more sense to just identity map all pages in 0xa0000 – 0xbffff as you proposed, at least for now. We could still add a config option or a cargo feature for this later so that it is possible to disable the mapping if desired (or maybe additionally identity-map other pages).

RKennedy9064:
Is that something you want me to add, or do you want to? Also if you have a some spare time, any chance you can look at the github action for https://github.com/rust-osdev/vga? I tried setting something similar up to how `x86_64` works and everything works except for the cargo xtest step, which is weird because I can run it locally. I haven't really worked with github actions much.

phil-opp:
I'd happy to merge a PR for this if you have time!

>  Also if you have a some spare time, any chance you can look at the github action for https://github.com/rust-osdev/vga? I tried setting something similar up to how `x86_64` works and everything works except for the cargo xtest step

I think the problem was that the project config file was named `cargo.toml` and not `Cargo.toml`. I opened https://github.com/rust-osdev/vga/pull/1 to correct this.

RKennedy9064:
This is now part of https://github.com/rust-osdev/vga

PR #105, opened 2020-04-03
SSE feature: remove inline assembly + don't set reserved bits

Freax13:
- get rid of inline assembly
- for some reason the previous version also set bits 9 & 10 in Cr0. according to the developer manual bits 5-16 are reserved

### DIFF
--- a/src/sse.rs
+++ b/src/sse.rs
@@ -1,23 +1,17 @@
 /// Enables Streaming SIMD Extensions (SSE) support for loaded kernels.
 pub fn enable_sse() {
-    use bit_field::BitField;
-    use x86_64::registers::control::Cr0;
-    let mut flags = Cr0::read_raw();
-    flags.set_bit(2, false);
-    flags.set_bit(1, true);
-    flags.set_bit(9, true);
-    flags.set_bit(10, true);
+    use x86_64::registers::control::{Cr0, Cr0Flags, Cr4, Cr4Flags};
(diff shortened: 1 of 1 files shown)


phil-opp:
Thanks a lot for the pull request!

I'm not sure why we currently set bits 9 and 10 in CR0. From a quick look at the manuals, it seems like you're right that these bits are reserved. Maybe this was a copy&paste error? Perhaps @ethindp who added this code knows more?

phil-opp:
Given that this change only affects the SSE feature and nobody objected it, I'm going to merge it. After all, setting bits 9 and 10 seems like an accident to me.

Thanks for the PR, @Freax13!

phil-opp:
Published as version 0.9.1

ethindp:
I'm honestly not sure myself. The OSDev Wiki listed the code I added
to enable SSE -- I didn't write it, though I probably should have (I
need ASM practice :)). Perhaps its worth asking on the OSDev Forum?

On 4/15/20, Philipp Oppermann <notifications@github.com> wrote:
> Published as version 0.9.1
>
> --
> You are receiving this because you were mentioned.
> https://github.com/rust-osdev/bootloader/pull/105#issuecomment-613858324


-- 
Signed,
Ethin D. Probst

phil-opp:
Thanks for clarifying! I think as long as it still works, setting these bits was likely a copy&paste error.

PR #108, opened 2020-04-25
Use `llvm_asm!` instead of deprecated `asm!`

### LINKED ISSUE #107: Deprecation: `asm!` > `llvm_asm!`
Should be a pretty simple problem to fix, but as of 24 April 2020, use of the `asm!` macro is triggering multiple deprecation warnings:

    warning: use of deprecated item 'asm': the syntax of asm! will change soon, use llvm_asm! to avoid breakage
      --> src/main.rs:46:5
       |
    46 |     asm!("call $1; ${:private}.spin.${:uid}: jmp ${:private}.spin.${:uid}" ::
       |     ^^^ help: replace the use of the deprecated item: `llvm_asm`
       |
       = note: `#[warn(deprecated)]` on by default
    
    warning: use of deprecated item 'asm': the syntax of asm! will change soon, use llvm_asm! to avoid breakage
      --> src/main.rs:92:5
       |
    92 |     asm!("mov bx, 0x0
       |     ^^^ help: replace the use of the deprecated item: `llvm_asm`

Should be an easy fix if you ask me ― since there's probably a lot more instances than those two, is this being worked on already?

realKennyStrawn93:
Fixes #107

This potentially **breaks users on older nightlies**.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,7 @@
 #![feature(lang_items)]
 #![feature(global_asm)]
 #![feature(step_trait)]
-#![feature(asm)]
+#![feature(llvm_asm)]
 #![feature(nll)]
 #![feature(const_fn)]
 #![no_std]
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,7 +43,7 @@ global_asm!(include_str!("video_mode/vga_320x200.s"));
 global_asm!(include_str!("video_mode/vga_text_80x25.s"));
 
 unsafe fn context_switch(boot_info: VirtAddr, entry_point: VirtAddr, stack_pointer: VirtAddr) -> ! {
-    asm!("call $1; ${:private}.spin.${:uid}: jmp ${:private}.spin.${:uid}" ::
+    llvm_asm!("call $1; ${:private}.spin.${:uid}: jmp ${:private}.spin.${:uid}" ::
          "{rsp}"(stack_pointer), "r"(entry_point), "{rdi}"(boot_info) :: "intel");
     ::core::hint::unreachable_unchecked()
 }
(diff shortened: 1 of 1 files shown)


phil-opp:
Published as v0.9.2

qiujiangkun:
```log
   Compiling bootloader v0.9.19 (/Users/jack/.cargo/registry/src/github.com-1ecc6299db9ec823/bootloader-0.9.19)
warning: use of deprecated macro `llvm_asm`: will be removed from the compiler, use asm! instead
  --> src/main.rs:45:5
   |
45 |     llvm_asm!("call $1; ${:private}.spin.${:uid}: jmp ${:private}.spin.${:uid}" ::
   |     ^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

warning: use of deprecated macro `llvm_asm`: will be removed from the compiler, use asm! instead
  --> src/main.rs:91:5
   |
91 |     llvm_asm!("mov bx, 0x0
   |     ^^^^^^^^

warning: `bootloader` (bin "bootloader") generated 2 warnings
```

Why is this happening?

bjorn3:
`llvm_asm!()` is now deprecated. In it's place a new `asm!()` syntax has been created. The original `asm!()` macro was renamed to `llvm_asm!()` before the new `asm!()` macro was introduced. Basically the migration path from the old to the new asm syntax is: `asm!()` renamed to `llvm_asm!()`. New `asm!()` introduced. `llvm_asm!()` deprecated. (current step) And then in the future removing `llvm_asm!()`. Bootloader will need to be updated to the new `asm!()` syntax before the old syntax is removed. This has already been done for the 0.10 series in https://github.com/rust-osdev/bootloader/pull/154, but the 0.9 series still uses `llvm_asm!()`.

PR #114, opened 2020-04-29
Rewrite - Arch subcrates

rybot666:
Add arch-specific subcrates that are staticlibs. Custom build script to link


.cargo/config
```diff
@@ -1,2 +1,5 @@
 [build]
-target = "x86_64-real_mode.json"
\ No newline at end of file
+target = "x86_64-real_mode.json"
+
+[alias]
+xbuild = "build -Zbuild-std=core"
```
bjorn3, line 5: nit: missing trailing newline

rybot666: Fixed for next commit

PR #115, opened 2020-04-30
Revert OUT_DIR output

rybot666:
Turns out a change I made makes it not build on my machine in particular (despite CI passing). I think this is due to cargo artifacts being deleted and used simultaneously (no backing for that though). Anyway, this fixes it

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,15 +8,17 @@ use std::process::Command;
 
 fn main() {
     // Read environment variables set by cargo
-    let out_dir_path = env::var("OUT_DIR").expect("Missing OUT_DIR environment variable");
-    let out_dir = Path::new(&out_dir_path);
-
     let cargo_path = env::var("CARGO").expect("Missing CARGO environment variable");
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -8,16 +8,18 @@ use std::process::Command;
...
     let cargo_path = env::var("CARGO").expect("Missing CARGO environment variable");
     let cargo = Path::new(&cargo_path);
 
     let manifest_dir_path =
         env::var("CARGO_MANIFEST_DIR").expect("Missing CARGO_MANIFEST_DIR environment variable");
     let manifest_dir = Path::new(&manifest_dir_path);
 
+    // Find the root project target dir
+    let current_dir = env::current_dir().expect("Couldn't get current directory");
+    let target_dir_rel = manifest_dir.join("target");
```
bjorn3, line 20: For crates.io dependencies I believe the current dir is the source dir of the dependency, not the source dir of the original project.

bjorn3: Ping @rybot666?

PR #117, opened 2020-05-20
Update x86_64 dependency to version 0.11.0

phil-opp:
This should fix the build error caused by the deprecated `asm!` macro on recent nightlies.


phil-opp:
Published as v0.9.3.

PR #123, opened 2020-07-07
Change 1st stage int 13h addressing

rsribeiro:
Changed the addressing of the bootloader loading from disk on stage 1, so the buffer doesn't cross the memory segment boundary, as is required by int 13h/ah=42h (see [Wikipedia](https://en.wikipedia.org/wiki/INT_13H#INT_13h_AH=42h:_Extended_Read_Sectors_From_Drive)).

With this change the bootloader no longer gets stuck on stage 1 on VirtualBox. Also tested on QEMU and Bochs, but not on real hardware.

Feel free to rewrite in a better way as I have no previous experience with asm.

### DIFF
--- a/src/stage_1.s
+++ b/src/stage_1.s
@@ -80,8 +80,17 @@ check_int13h_extensions:
 load_rest_of_bootloader_from_disk:
     lea eax, _rest_of_bootloader_start_addr
 
-    # start of memory buffer
+    # dap buffer segment
+    mov ebx, eax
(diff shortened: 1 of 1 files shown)


phil-opp:
Thanks a lot for this! Seems like I missed the fact that the buffer must be inside the same memory segment. This also means that our bootloader must not be larger than 64KiB right?

The implementation looks good to me!

phil-opp:
Published as v0.9.6.

PR #125, opened 2020-07-17
Make bootloader buildable with `-Zbuild-std`

phil-opp:
This switches the bootloader build from `cargo-xbuild` to `cargo build -Zbuild-std=core`. It uses the new `package.metadata.bootloader.build-std` key of version 0.8.1 of `bootimage` for that.

This change is non-breaking since the bootloader continues to be buildable with `cargo-xbuild`.


phil-opp:
Published as v0.9.7.

PR #126, opened 2020-07-17
Enable rlibc dependency only with `binary` feature

phil-opp:
The dependency is not needed when compiling the bootloader crate as a library.


phil-opp:
Published as v0.9.8.

PR #129, opened 2020-09-29
Update Cargo.lock to prevent nightly breakage

josephlr:
See https://github.com/rust-osdev/x86_64/pull/186#issuecomment-700491664

Now that x86_64 has a "good" build, we just need a new lockfile to make things build.

Signed-off-by: Joe Richey <joerichey@google.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,9 +80,9 @@ checksum = "f70329e2cbe45d6c97a5112daad40c34cd9a4e18edb5a2a18fefeb584d8d25e5"
 
 [[package]]
 name = "x86_64"
-version = "0.12.1"
+version = "0.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "238a5798f77641af3c4f242bf985807f312a480cd4e35ed7255fad4b2ccb9d4f"
+checksum = "d5b4b42dbabe13b69023e1a1407d395f1a1a33df76e9a9efdbe303acc907e292"
 dependencies = [
  "bit_field 0.9.0",
  "bitflags",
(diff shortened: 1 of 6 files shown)


phil-opp:
Thanks! Could you also update the Cargo.lock files in `example-kernel` and `test-kernel` (these are also built in our current test framework).

josephlr:
@phil-opp This needs to be force merged

> Thanks! Could you also update the Cargo.lock files in `example-kernel` and `test-kernel` (these are also built in our current test framework).

Done

phil-opp (approved):
Thanks!

josephlr:
> Thanks!

@phil-opp It looks like one of the tests got (inadvertently) cancelled. I don't have permissions to rerun them.

phil-opp:
Seems like the Windows job is having some problems downloading QEMU. I already restarted it once after seeing the same issue. Let's try it again…

josephlr:
> Seems like the Windows job is having some problems downloading QEMU. I already restarted it once after seeing the same issue. Let's try it again…

Looks like a network failure. The Windows test consistently fails in downloading (the link works just fine for me locally though)

phil-opp:
I also just tried it locally and it worked for me too. I'm not sure what we should do about this. I'm inclined to force-merge and publish to unblock the nightly features, but on the other hand it feels not good to publish a new version without a passing CI.

josephlr:
> I also just tried it locally and it worked for me too. I'm not sure what we should do about this. I'm inclined to force-merge and publish to unblock the nightly features, but on the other hand it feels not good to publish a new version without a passing CI.

It's building and all the other tests and stuff are passing. The downside here is that without this change, the CI's for all the other projects aren't going to pass. Given that this would be the _only_ change for `0.9.11`. I think force merging is OK.

phil-opp:
Thanks! Ok, then let's do this.

phil-opp:
Published as v0.9.11

PR #134, opened 2020-12-10
Documentation

Luis-Hebendanz:
I documented the build process and added `sse` to the feature list and a little debug help section

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,77 @@ Written for the [second edition](https://github.com/phil-opp/blog_os/issues/360)
 
 ## Design
 
-TODO
+When you press the power button the computer loads the BIOS from some flash memory stored on the motherboard.
+The BIOS initializes and self tests the hardware then loads the first 512 bytes into memory from the media device
+(i.e. the cdrom or floppy disk). If the last two bytes equal 0xAA55 then the BIOS will jump to location 0x7C00 effectively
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,13 +157,28 @@ dd if=target/x86_64-bootloader/release/bootloader.bin of=/dev/sdX && sync
 
 Where sdX is the device name of your USB stick. **Be careful** to choose the correct device name, because everything on that device is overwritten.
 
+## Debugging
+Set a breakpoint at address `0x7c00`. Disassemble instructions with gdb:
```
bjorn3, line 161: This address will differ between versions of bootloader and possibly between assembler versions.

rybot666: That's the address BIOS jumps to to start code execution, we'll always have *something* there

bjorn3: Ah, ok.

ethindp:
I disagree; when the bootloader gains UEFI support 0x7C00 will no longer be a constant address, I believe.

Luis-Hebendanz:
> I disagree; when the bootloader gains UEFI support 0x7C00 will no longer be a constant address, I believe.

I mean, the documentation documents the current behavior if this changes the documentation should be changed accordingly

phil-opp (approved):
Looks good to me, thanks a lot!

I agree that we can leave the `0x7c00` debug address in for now, we will have to update the documentation in many places for the new UEFI bootloader anyway.

PR #135, opened 2021-02-02
Fix build on latest nightly by updating x86_64 to v0.13.2

phil-opp:


phil-opp:
Merging without review to fix nightly failure.

phil-opp:
Published as v0.9.12

PR #139, opened 2021-03-07
Fix linker errors on latest nightlies

phil-opp:
Don't use the `lea` instructions get addresses of non-addressable memory locations. Instead, use `mov <reg>, offset <label>` instructions, which are the equivalent to `mov <reg>, label` instructions in `nasm`.

I'm not sure why the `lea` worked before, but maybe LLVM used to optimize the `lea`s to `mov` instructions. Either way, the LLVM version used by the latest Rust nightly doesn't like the `lea` instructions anymore.

This PR includes a lot of commits where I tried to fix the Windows CI, which currently fails because of a [bug in QEMU](https://bugs.launchpad.net/qemu/+bug/1915794). Unfortunately, I wasn't able to get it working, so I just hope for a new QEMU release now. For the meantime, I disabled the `fail-fast` property so that we at least don't cancel the macOS build when the Windows job fails. (I kept the debug commits in this PR in case so that I don't need to rember which approaches I tried the next time I debug this.)

Fixes https://github.com/rust-osdev/bootimage/issues/73 and fixes https://github.com/phil-opp/blog_os/issues/937

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,7 @@ jobs:
         HOMEBREW_NO_INSTALL_CLEANUP: 1
     - name: Install QEMU (Windows)
       run: |
-        choco install qemu --no-progress
+        choco install qemu --version=2021.02.03 -y --no-progress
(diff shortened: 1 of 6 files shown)


phil-opp:
Published as v0.9.15.

PR #140, opened 2021-03-07
Replace all remaining `lea`s with `mov` + `offset`

phil-opp:
Follow-up to #139 since there were still [some relocation errors](https://github.com/phil-opp/blog_os/issues/937#issuecomment-792306992) for real-world kernels.


phil-opp:
Published as v0.9.16

PR #147, opened 2021-04-07
Fix documented type for entry point function

mwkroening:
What do you think?

### DIFF
--- a/src/boot_info.rs
+++ b/src/boot_info.rs
@@ -8,7 +8,7 @@ use core::{ops, slice};
 /// ```
 /// # use bootloader::BootInfo;
 /// # type _SIGNATURE =
-/// extern "C" fn(boot_info: &'static BootInfo) -> !;
+/// extern "C" fn(boot_info: &'static mut BootInfo) -> !;
 /// ```
 ///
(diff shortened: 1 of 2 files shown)


src/boot_info.rs
```diff
@@ -8,7 +8,7 @@ use core::{ops, slice};
 /// ```
 /// # use bootloader::BootInfo;
 /// # type _SIGNATURE =
-/// extern "C" fn(boot_info: &'static BootInfo) -> !;
+/// fn(boot_info: &'static mut BootInfo) -> !;
```
phil-opp, line 11: The `extern "C"` is still part of the signature, even if it's not needed with the `entry_point` macro. So I think we should keep it here and instead add to the docs of `entry_point!` that it is not needed when using the macro.

phil-opp (commented):
Good catch, thanks a lot!

src/boot_info.rs
mwkroening: You are totally correct, thanks!

The docs of the macro don't mention `extern "C"`, so I think it's fine. :thinking:

phil-opp:
Thanks!

PR #154, opened 2021-04-27
llvm_asm! -> asm!

CraftSpider:
Tiny change to stop using `llvm_asm!`, as everywhere else is using `asm!` and the feature is basically deprecated.


phil-opp:
Thanks a lot!

phil-opp:
Published together with #155 as v0.10.2.

qiujiangkun:
Could you backport it to v0.9?

PR #155, opened 2021-04-30
Reduce the number of used unstable features of x86_64 crate

phil-opp:
The `const_fn` feature breaks quite often, e.g. most recently in https://github.com/rust-lang/rust/pull/84310. We don't need this kind of const_fn functionality for this crate, so it makes sense to disable it to avoid such breakage.

### DIFF
--- a/tests/test_kernels/default_settings/Cargo.toml
+++ b/tests/test_kernels/default_settings/Cargo.toml
@@ -6,5 +6,5 @@ edition = "2018"
 
 [dependencies]
 bootloader = { path = "../../.." }
-x86_64 = { version = "0.13.2", optional = true, default-features = false, features = ["instructions", "inline_asm"] }
+x86_64 = { version = "0.13.2", default-features = false, features = ["instructions", "inline_asm"] }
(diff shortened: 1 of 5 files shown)


tests/test_kernels/default_settings/Cargo.toml
```diff
@@ -6,5 +6,5 @@ edition = "2018"
 
 [dependencies]
 bootloader = { path = "../../.." }
-x86_64 = "0.13.2"
+x86_64 = { version = "0.13.2", optional = true, default-features = false, features = ["instructions", "inline_asm"] }
```
bjorn3, line 9:
```suggestion
x86_64 = { version = "0.13.2", default-features = false, features = ["instructions", "inline_asm"] }
```
This wasn't optional before. Same for the other test kernel.

phil-opp: Good catch, thanks a lot!

phil-opp:
Published together with #154 as v0.10.2.

phil-opp:
Backported to v0.9 as `v0.9.17`.

PR #156, opened 2021-05-04
Change register used in setting SS in stage_4

Elekrisk:
The register `bx` is apparently reserved by LLVM, and building with the latest nightly (1.54-nightly) fails because of it:

```
error: invalid register `bx`: rbx is used internally by LLVM and cannot be used as an operand for inline asm
  --> src\bin\bios.rs:55:9
   |
55 |         out("bx") _,
   |  
```

This pull request changes the register used to `ax`.
I don't think this'll be able to break anything. `cargo test` passes, and building and running my own kernel works.

### DIFF
--- a/src/bin/bios.rs
+++ b/src/bin/bios.rs
@@ -51,8 +51,8 @@ extern "C" {
 pub unsafe extern "C" fn stage_4() -> ! {
     // Set stack segment
     asm!(
-        "mov bx, 0x0; mov ss, bx",
-        out("bx") _,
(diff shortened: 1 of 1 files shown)


phil-opp:
Thanks for the pull request! I just restarted our CI job in order to test it with the latest nightly and I can confirm that the mentioned error occurs. It looks like it is caused by https://github.com/rust-lang/rust/pull/84658. So apparently using `rbx` was already incorrect before, it just didn't lead to an error.

phil-opp:
Published as v0.10.3.

Issue #159, opened 2021-05-13
Higher Half?


Andy-Python-Programmer:
Will be nice to support higher half kernels.

phil-opp:
Have you tried to specify a custom virtual base address via an `--image-base <addr>` argument when building your kernel? Alternatively, you can use a linker script for this. The bootloader should be able to load the kernel into the virtual address space at exactly the specified address, which should also be possible with addresses in the higher half. I say "should" because I have not tested this yet since the UEFI rewrite. Please let me know if it doesn't work so that we can fix it!

Andy-Python-Programmer:
> Alternatively, you can use a linker script for this. 
I used the following linker script and is verified on the limine bootloader that it works fine and the higher offset value is fine.

```ld
ENTRY(_start)
OUTPUT_FORMAT(elf64-x86-64)

KERNEL_OFFSET = 0xFFFF800000100000;

SECTIONS {
    . = KERNEL_OFFSET;

    . += SIZEOF_HEADERS;
    . = ALIGN(4096);

    .text : AT(ADDR(.text) - KERNEL_OFFSET) {
        __text_start = .;
        *(.text*)
        . = ALIGN(4096);
        __text_end = .;
    }

    .rodata : AT(ADDR(.rodata) - KERNEL_OFFSET) {
        __rodata_start = .;
        *(.rodata*)
        . = ALIGN(4096);
        __rodata_end = .;
    }

    .data : AT(ADDR(.data) - KERNEL_OFFSET) {
        __data_start = .;
        *(.data*)
        . = ALIGN(4096);
        __data_end = .;
        __bss_start = .;
        *(.bss*)
        . = ALIGN(4096);
        __bss_end = .;
    }

    .tdata : AT(ADDR(.tdata) - KERNEL_OFFSET) {
        __tdata_start = .;
        *(.tdata*)
        . = ALIGN(4096);
        __tdata_end = .;
        __tbss_start = .;
        *(.tbss*)
        . += 8;
        . = ALIGN(4096);
        __tbss_end = .;
    }

    __kernel_end = .;

    /DISCARD/ : {
        *(.comment*)
        *(.eh_frame*)
        *(.gcc_except_table*)
        *(.note*)
        *(.rel.eh_frame*)
    }
}
```

![image](https://user-images.githubusercontent.com/62820092/118228886-50652280-b4ce-11eb-8bb1-7f822909794b.png)

[...]

PR #160, opened 2021-05-14
Reserve 3 frames for higher half kernels

Andy-Python-Programmer:
This pull request reserves three frames instead of two for higher half kernel support.

Related: #159

### DIFF
--- a/src/binary/mod.rs
+++ b/src/binary/mod.rs
@@ -262,7 +262,7 @@ pub fn create_boot_info<I, D>(
     page_tables: &mut PageTables,
     mappings: &mut Mappings,
     system_info: SystemInfo,
-) -> (&'static mut BootInfo, TwoFrames)
+) -> (&'static mut BootInfo, ReservedFrames)
 where
     I: ExactSizeIterator<Item = D> + Clone,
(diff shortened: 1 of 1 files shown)


phil-opp:
Thanks for the PR! I'm currently looking into this issue too and I think there might be a better way to solve this. I'll keep you posted.

Andy-Python-Programmer:
All good!

Andy-Python-Programmer:
For reference the limine bootloader subtracts the fixed higher half offset from the elf header section.

<https://github.com/limine-bootloader/limine/blob/f650e4b41bda371dd8d6fd764bc2c77bcbc3f102/stage23/lib/elf.c#L339>

Andy-Python-Programmer:
Closing as of #161

phil-opp:
I opened #161 with an alternative fix.



> For reference the limine bootloader subtracts the fixed higher half offset from the elf header section.

I'm not sure why this is done, maybe because the bootloader is still in 32 bit mode at this point? In our case, we are already in 64 bit mode when doing the mapping, so we can create mappings across the whole virtual address space without problems.

PR #165, opened 2021-05-16
Update dependencies to fix compilation errors

ethindp:
This PR fixes a compilation error with bootloader v 0.10.4 which relies on x86_64 0.13.2. On Rust nightly (at least on rustc 1.54.0-nightly (8cf990c9b 2021-05-15) and later), this version uses removed features. This PR updates x86_64 to version 0.14.2 to fix this problem.


ethindp:
I should also note that this updates all other outdated dependencies as well. All tests passed and the bootloader still builds fine.

phil-opp:
Thanks for the PR! Could you clarify which error occurs for you? I just tried reproducing it with the same nightly but `cargo test` seems to run fine. (I just try to understand what kind of breakage occurs so that we can react accordingly.)

ethindp:
Understandable! Specifically, the error that's caused is E0557. You use const_fn and friends that were removed in, I believe, rust 1.53 nightly, and replaced with more fine-grained features, in x86_64 0.13. At least, I'm getting that error, and so I assumed it was the bootloader causing it since my kernel doesn't use v. 0.13.6 of the x86_64 crate but 0.14.2.

phil-opp:
Thanks for clarifying! Have you tried running `cargo update`? I think we backported the fix for this issue to v0.13, but I'm not completely sure.

ethindp:
Yeah, I'm on the latest deps and it still gives me that error.

On 5/16/21, Philipp Oppermann ***@***.***> wrote:
> Thanks for clarifying! Have you tried running `cargo update`? I think we
> backported the fix for this issue to v0.13, but I'm not completely sure.
>
> --
> You are receiving this because you authored the thread.
> https://github.com/rust-osdev/bootloader/pull/165#issuecomment-841844397


-- 
Signed,
Ethin D. Probst

ethindp:
Okay, so I figured out why. It wasn't the bootloader crate, it was the
vga crate. I'm not even sure if I need that or if I should just use
something like fontdue and use a TTF font to render text.

On 5/16/21, Ethin Probst ***@***.***> wrote:
> Yeah, I'm on the latest deps and it still gives me that error.
>
> On 5/16/21, Philipp Oppermann ***@***.***> wrote:
>> Thanks for clarifying! Have you tried running `cargo update`? I think we
>> backported the fix for this issue to v0.13, but I'm not completely sure.
>>
>> --
>> You are receiving this because you authored the thread.
>> Reply to this email directly or view it on GitHub:
>> https://github.com/rust-osdev/bootloader/pull/165#issuecomment-841844397
>
>
> --
> Signed,
> Ethin D. Probst
>


-- 
Signed,
Ethin D. Probst

phil-opp:
Ah, good to hear, thanks!

phil-opp:
Looks like the issue was solved, so I think we can close this PR.

PR #168, opened 2021-05-20
Fix nightly regression by manually passing `--gc-sections`

### LINKED ISSUE #167: Bootloader `v0.9.17` seems to be broken on Rust `nightly-2021-05-20`
I just noticed this in the automated CI runs for the `blog_os` project: https://github.com/phil-opp/blog_os/runs/2626134888?check_suite_focus=true#step:12:47

When trying to run `cargo +nightly-2021-05-20 run`, the following error occurs:

image

The error message is "Failed to load rest of bootloader" and happens in the early assembly stages.

I'm not sure what's going wrong yet, I will update this issue when I find the cause.

phil-opp:
Rust used to pass `--gc-section` automatically. In https://github.com/rust-lang/rust/pull/85274, this behavior was changed to only pass that flag for targets that use a GNU linker. So we have to add it manually, otherwise the bootloader grows too large to be loaded by our assembly stage.

Fixes #167

This is a hotfix for the `v0.9.17` release and does not apply to the `main` branch (which needs some additional fixes).

### DIFF
--- a/x86_64-bootloader.json
+++ b/x86_64-bootloader.json
@@ -5,9 +5,9 @@
     "linker": "rust-lld",
     "pre-link-args": {
         "ld.lld": [
-	        "--script=linker.ld",
+            "--script=linker.ld",
             "--gc-sections"
-	    ]
+        ]
     },
     "target-endian": "little",
     "target-pointer-width": "64",
--- a/x86_64-bootloader.json
+++ b/x86_64-bootloader.json
@@ -18,5 +18,5 @@
     "disable-redzone": true,
     "panic": "abort",
     "executables": true,
-	"relocation_model": "static"
+    "relocation_model": "static"
 }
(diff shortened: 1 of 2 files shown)


x86_64-bootloader.json
```diff
@@ -5,8 +5,9 @@
     "linker": "rust-lld",
     "pre-link-args": {
         "ld.lld": [
-	    "--script=linker.ld"
-	]
+	        "--script=linker.ld",
+            "--gc-sections"
+	    ]
```
bjorn3, line 10:
```suggestion
            "--script=linker.ld",
            "--gc-sections"
        ]
```
replacing tabs with spaces

phil-opp: Thanks!

phil-opp:
Published as v0.9.18

PR #174, opened 2021-05-24
Uefi: Look for an ACPI2 RSDP first

phil-opp:
Only look for an ACPI v1 RSDP if no v2 RSDP was found.

Fixes https://github.com/rust-osdev/bootloader/issues/172.

PR #176, opened 2021-05-24
Don't check target architecture for builder crate to support cross-compiling

### LINKED ISSUE #173: Support cross-compiling in `builder`
I am working on my operating system on an Apple Silicon M1 MacBook (aarch64) from time to time and when I try to build the boot image, it always fails with the message `This crate only supports the x86_64 architecture.` because it uses the current architecture, of course. I ended up cloning the repo and adding the `TARGET` environment variable to the builder process to make it cross-compile for x86_64, which works for me but I'm not sure if that's the best way to do this.

Maybe the crate could get a config option for cross-compiling or something? Or always force a specific target? I'm not sure what issues might arise from this, so I thought I'd submit this issue first to discuss stuff.

If we can agree on a good solution, I'd be happy to submit a PR for that! :)

Thanks for your work!

phil-opp:
Fixes #173

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 - Identity-map GDT into kernel address space to fix `iretq` ([#175](https://github.com/rust-osdev/bootloader/pull/175))
 - Uefi: Look for an ACPI2 RSDP first ([#174](https://github.com/rust-osdev/bootloader/pull/174))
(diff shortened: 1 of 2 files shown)


phil-opp:
Published together with #175 and #174 as `v0.10.6`.

Issue #164, opened 2021-05-15
Pass `UnwindInfo` in `BootInfo`


Andy-Python-Programmer:
It will be nice to pass `UnwindInfo` in `BootInfo` so that we can load the kernel ELF in the kernel to do stack unwinding.

```rust
#[derive(Debug, Copy, Clone)]
#[repr(C)]
pub struct UnwindInfo {
    /// The base address of the kernel. The kernel base is required
    /// for stack unwinding during kernel panics.
    pub kernel_base: VirtAddr,
    /// The size of the kernel, required to calculate the end of the
    /// kernel base.
    pub kernel_size: usize,

    pub stack_top: VirtAddr,
    pub stack_size: usize
}
```

phil-opp:
Sounds reasonable to me! While you can already find out the kernel start address and size through a linker script, I agree that a boot info field would be easier. For stack top and size there is currently no workaround except for setting these to fixed values in the bootloader config.

ethindp:
@phil-opp Couldn't you just pass that after finding the address in the bootloader? I'm confused about the problem.

phil-opp:
There is no problem here, we just need to do it. Happy to merge a PR for this!

phil-opp:
Some adjustments to the code from the OP that I would make:

```rust
#[derive(Debug, Copy, Clone)]
#[repr(C)]
pub struct KernelInfo {
    pub kernel_base: u64
    pub kernel_size: u64,
    pub stack_top: u64,
    pub stack_size: u64,
}
```

Basically:

- Name the struct `KernelInfo` instead of `UnwindInfo` since it isn't directly related to unwinding and might have other use cases too.
- Don't use the `VirtAddr` type of the `x86_64` crate to keep the `BootInfo` stable (otherwise updating the `x86_64` version would be a breaking change).
- Don't use platform dependent size types such as `usize`. Use `u64` instead. (This might be overcautious, but it makes sure that the `BootInfo` always has the same size independent of the target platform.)

PR #177, opened 2021-05-25
Add support for kernel info

Andy-Python-Programmer:
Fixes: #164

### DIFF
--- a/src/binary/mod.rs
+++ b/src/binary/mod.rs
@@ -1,6 +1,6 @@
 use crate::{
     binary::legacy_memory_region::{LegacyFrameAllocator, LegacyMemoryRegion},
-    boot_info::{BootInfo, FrameBuffer, FrameBufferInfo, MemoryRegion, TlsTemplate},
+    boot_info::{BootInfo, FrameBuffer, FrameBufferInfo, KernelInfo, MemoryRegion, TlsTemplate},
 };
 use core::{
     mem::{self, MaybeUninit},
@@ -139,8 +139,9 @@ where
     // create a stack
     let stack_start_addr = kernel_stack_start_location(&mut used_entries);
     let stack_start: Page = Page::containing_address(stack_start_addr);
+    let stack_size = CONFIG.kernel_stack_size.unwrap_or(20 * PAGE_SIZE);
     let stack_end = {
-        let end_addr = stack_start_addr + CONFIG.kernel_stack_size.unwrap_or(20 * PAGE_SIZE);
+        let end_addr = stack_start_addr + stack_size;
         Page::containing_address(end_addr - 1u64)
     };
     for page in Page::range_inclusive(stack_start, stack_end) {
@@ -259,6 +260,13 @@ where
         None
     };
 
+    let kernel_info = KernelInfo {
+        kernel_base: kernel_bytes.as_ptr() as u64,
+        kernel_size: kernel_bytes.len() as u64,
+        stack_top: stack_end.start_address().as_u64(),
+        stack_size,
+    };
+
     Mappings {
         framebuffer: framebuffer_virt_addr,
         entry_point,
@@ -267,6 +275,7 @@ where
         physical_memory_offset,
         recursive_index,
         tls_template,
+        kernel_info,
     }
 }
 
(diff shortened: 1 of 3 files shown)


Andy-Python-Programmer:
cc @phil-opp?

src/binary/mod.rs
```diff
@@ -246,6 +247,13 @@ where
         None
     };
 
+    let kernel_info = KernelInfo {
+        kernel_base: kernel_bytes.as_ptr() as u64,
```
phil-opp, line 251: This is the physical memory address of the kernel ELF file. To get the virtual address at which the kernel is mapped, we need to look at the segments in the program header of the ELF file. Note that there is no guarantee that these segments are continuous in memory since each section can be mapped arbitrarily using a linker script.

src/binary/mod.rs
```diff
@@ -246,6 +247,13 @@ where
...
+        kernel_size: kernel_bytes.len() as u64,
```
phil-opp, line 252: This is the size of the kernel ELF file, not the size that the kernel occupies in memory. Also, as noted in the previous comment, the kernel segments can be spread across the whole virtual address space, so I'm not sure if we can even choose a proper `kernel_size` for all kernels.

phil-opp (commented):
Thanks for the pull request and sorry for the late review!

Thinking more about this, I'm not sure if the `kernel_base` and `kernel_size` fields really make sense since not all kernels are continous in virtual memory. Perhaps you can clarify how these fields are required for implementing unwinding? Perhaps we can find a better solution.

Andy-Python-Programmer:
> Thinking more about this, I'm not sure if the kernel_base and kernel_size fields really make sense since not all kernels are continous in virtual memory. Perhaps you can clarify how these fields are required for implementing unwinding? Perhaps we can find a better solution.

Basically in stack unwinding we load the kernel bytes from the provided kernel_base and kernel_size then just `ElfFile::new(kernel_slice)` and then we walk through the RBP and

```rust
for data in symbol_table { // symbol table got from the symtab
       let st_value = data.value() as usize;
       let st_size = data.size() as usize;

       if rip >= st_value && rip < (st_value + st_size) { 
           let mangled_name = data.get_name(&kernel_elf);
         ... 
       }
} // basically what you could do using readelf and include it using `include_bytes` but this is runtime thinggy
 ```

Andy-Python-Programmer:
@phil-opp a better alternative solution would be to pass the symbol tab (symtab) section in the KernelInfo itself as we are loading the ELF in the bootloader right (so we would save an ELF load on unwind)?

ethindp:
Wouldn't both strategies fail in release mode? You don't have
debugging information in release mode, which, if I remember right,
means you have no symbol table either.

On 6/16/21, Anhad Singh ***@***.***> wrote:
> @phil-opp a better alternative solution would be to pass the symbol tab
> (symtab) section in the KernelInfo itself as we are loading the ELF in the
> bootloader right (so we would save an ELF load on unwind)?
>
> --
> You are receiving this because you are subscribed to this thread.
> https://github.com/rust-osdev/bootloader/pull/177#issuecomment-862131983


-- 
Signed,
Ethin D. Probst

phil-opp:
@Andy-Python-Programmer Sorry for the delay, the past two months were quite busy for me.

> Basically in stack unwinding we load the kernel bytes from the provided kernel_base and kernel_size then just `ElfFile::new(kernel_slice)`

Thanks for the explanation, I think I understand it now. However, I'm not sure if it would work as implemented. The problem is that I'm not sure if we keep the unused (i.e. not loaded) parts of the ELF file reserved in physical memory after switching to the kernel. And even if we do currently, we might want to optimize this in the future.

So it would probably be better to add a config option for keeping the ELF file in physical memory after loading (similar to the existing config options). Only if it is enabled, we could fill an `Optional` `kernel_elf_address` field with the _physical_ address range of the ELF file (and reserve it in the physical memory map if we don't do it already).

The `stack_top` and `stack_size` field are independent of this and can be always set. So they should probably go into a different field. We should also note in the docs that these describe a _virtual_ address range.

phil-opp:
There is a new PR that implements something similar: https://github.com/rust-osdev/bootloader/pull/346

PR #179, opened 2021-06-20
UEFI GOP mode configuration

anellie:
This pull request adds `desired_framebuffer_height` and `desired_framebuffer_width` to the bootloader configuration, and makes the UEFI bootloader select and set a GOP mode with those parameters if one is available.

Some things I'm not quite happy with yet, but am unsure how to fix:
- This does handle BIOS/VESA. I sadly don't have much knowledge of x86 assembly and don't think I'd be able to add this
- `Config` gets written twice now, so that ` bin/uefi.rs` has access to the config. Seems not ideal, but unsure how to improve this.

This is my first PR and I hope it is useful. Thank you for this great project! It has been a great help in learning about OS development.

### DIFF
--- a/src/bin/uefi.rs
+++ b/src/bin/uefi.rs
@@ -7,6 +7,8 @@
 
 // Defines the constants `KERNEL_BYTES` (array of `u8`) and `KERNEL_SIZE` (`usize`).
 include!(concat!(env!("OUT_DIR"), "/kernel_info.rs"));
+// Contains the bootloader configuration specified by the kernel crate (needed for GOP configuration)
+include!(concat!(env!("OUT_DIR"), "/kernel_bootloader_config.rs"));
 
 static KERNEL: PageAligned<[u8; KERNEL_SIZE]> = PageAligned(KERNEL_BYTES);
 
@@ -18,10 +20,12 @@ use bootloader::{
     boot_info::FrameBufferInfo,
 };
 use core::{mem, panic::PanicInfo, slice};
+use parsed_config::CONFIG;
 use uefi::{
     prelude::{entry, Boot, Handle, ResultExt, Status, SystemTable},
     proto::console::gop::{GraphicsOutput, PixelFormat},
     table::boot::{MemoryDescriptor, MemoryType},
+    Completion,
 };
 use x86_64::{
     structures::paging::{FrameAllocator, OffsetPageTable, PageTable, PhysFrame, Size4KiB},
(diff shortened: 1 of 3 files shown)


src/bin/uefi.rs
```diff
@@ -149,6 +153,25 @@ fn init_logger(st: &SystemTable<Boot>) -> (PhysAddr, FrameBufferInfo) {
...
+                .filter(|m| m.info().resolution() == (width, height))
+                .last(),
+            (Some(height), None) => modes.filter(|m| m.info().resolution().1 == height).last(),
+            (None, Some(width)) => modes.filter(|m| m.info().resolution().0 == width).last(),
+            _ => None,
+        }
+    };
+    if let Some(mode) = mode {
+        gop.set_mode(&mode)
+            .expect_success("failed to apply gop mode");
```
Andy-Python-Programmer, line 172: Might be better to replace this panic message with `Failed to apply the desired display mode`.

Andy-Python-Programmer (commented):
For the BIOS video mode, the crate actually does set it up from what I see, I personally have not tried it but you can see it right here: https://github.com/rust-osdev/bootloader/blob/bc6c1bd62de697dc090ff8e7131cc36b91a531f4/src/asm/vesa.s#L131. Just from a quick view.

src/bin/uefi.rs
anellie: You're right, that's probably clearer for users of the crate. Fixed

anellie:
Read through `vesa.s` a bit more throughly and experimented a bit, VESA is also configurable now!

Andy-Python-Programmer:
No worries, mate!

Andy-Python-Programmer (approved):

anellie:
Just updated the PR to current `main`. @phil-opp would it be possible to get some feedback on this and hopefully merge it?

phil-opp:
Sorry for the delay and thanks a lot for tackling this!

Looks good overall. Some ideas for improvements:

- It looks like you're trying to get an exact resolution match on UEFI, but only use the supplied resolution as minimum bounds on BIOS. This seems a bit inconsistent. The easiest solution would be to set treat the numbers as minima for UEFI too (and change the config field names accordingly).
- The only reason for the duplicated config write is that you need different import paths for the `Config` struct, right? Perhaps it is possible to instead make the `parsed_config` module public and then access it directly from `bin/uefi.rs` (via `use bootloader::binary::parsed_config::CONFIG`) instead of including the config there too?

anellie:
Alright, got both of them done! UEFI now correctly treats it as minima (+ rename to `minimum_framebuffer_*`) and your idea of accessing the config module worked perfectly. Updated to current `main` as well.

anellie:
@phil-opp Any updates? This is ready for merge.

phil-opp (commented):
Thanks for the ping! The code looks good to me, thanks a lot! Have you tested both the BIOS and UEFI variants in QEMU?

It seems like the CI did not start for some reason. Could you push an additional dummy commit (or rebase the PR) to trigger it?

anellie:
I pushed another commit, however GitHub is telling me that the CI workflow is awaiting approval ("First-time contributors need a maintainer to approve running workflows", it links to [this article](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)).
Could you approve it? @phil-opp

PR #182, opened 2021-07-07
fix: also check cfg gated target field for bootloader dependency

### LINKED ISSUE #181: no bootloader dependency in `Cargo.toml` error when compiling, but it is there
I'm getting this error while compiling `bootloader`:
```
error: no bootloader dependency in /home/patriot/Belgeler/projects/hakkero/Cargo.toml

  The `--kernel-manifest` path should point to the `Cargo.toml` of the kernel.
 --> /home/patriot/Belgeler/projects/hakkero/target/x86_64-unknown-uefi/release/build/bootloader-b9ead764961ec1c9/out/bootloader_config.rs:1:81
  |
1 | ... = compile_error ! ("no bootloader dependency in /home/patriot/Belgeler/projects/hakkero/Cargo.toml\n\n  The `--kernel-manifest` path should point to the `Cargo.toml` of the kernel.") ...
  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: could not compile `bootloader`
```
it says I don't have a `bootloader` dependency in my `Cargo.toml`, but I do:
```
[target.'cfg(target_arch = "x86_64")'.dependencies]
bootloader = "0.10"
```

This is probably related to the handling in https://github.com/rust-osdev/bootloader/blob/c09f94f1fe96e21909437c7cfec33f5b4bb449fa/build.rs#L243-L246. This should be changed to handle cases like that. Maybe we could use https://github.com/oli-obk/cargo_metadata instead? If we don't want to pull in a dependency, it should also be possible to recursively travel through `dependencies`.

yusdacra:
This fixes #181 for me.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -242,7 +242,15 @@ mod binary {
 
                 if manifest
                     .get("dependencies")
-                    .and_then(|d| d.get("bootloader"))
+                    .and_then(|d| {
+                        d.get("bootloader").or_else(|| {
+                            manifest
+                                .get("target")
+                                .and_then(|table| table.get(r#"cfg(target_arch = "x86_64")"#))
+                                .and_then(|table| table.get("dependencies"))
+                                .and_then(|table| table.get("bootloader"))
+                        })
+                    })
                     .is_some()
                 {
                     // it seems to be the correct Cargo.toml
--- a/build.rs
+++ b/build.rs
@@ -242,14 +242,13 @@ mod binary {
 
                 if manifest
                     .get("dependencies")
-                    .and_then(|d| {
-                        d.get("bootloader").or_else(|| {
-                            manifest
-                                .get("target")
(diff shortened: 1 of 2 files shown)


phil-opp:
Thanks a lot and sorry for the delay!

Seems like the CI wasn't triggered somehow, let me try to fix this.

phil-opp:
Published together with https://github.com/rust-osdev/bootloader/pull/180 as v0.10.8

PR #185, opened 2021-07-16
Fix compilation warnings on v0.9 branch

vinc:
Hello, I get some warnings on the 0.9.18 version when compiling MOROS (https://github.com/vinc/moros)
```
warning: avoid using `.intel_syntax`, Intel syntax is the default
```
And also
```
warning: target json file contains unused fields: panic, relocation_model
```

The fixes for `intel_syntax` and `panic` already exist for the main branch so I adapted them.

I also removed `relocation_model` from the JSON target file and tested it on my OS with Bosch, QEMU, and VirtualBox but not yet on real hardware so I'm unsure about that one.

Edit: I renamed the later to `relocation-model` to fix the warning thanks to @bjorn3 (see https://doc.rust-lang.org/rustc/codegen-options/index.html#relocation-model)

### DIFF
--- a/x86_64-bootloader.json
+++ b/x86_64-bootloader.json
@@ -17,6 +17,5 @@
     "features": "-mmx,-sse,+soft-float",
     "disable-redzone": true,
     "panic-strategy": "abort",
-    "executables": true,
-    "relocation_model": "static"
+    "executables": true
 }
--- a/x86_64-bootloader.json
+++ b/x86_64-bootloader.json
@@ -17,5 +17,6 @@
     "features": "-mmx,-sse,+soft-float",
     "disable-redzone": true,
     "panic-strategy": "abort",
-    "executables": true
+    "executables": true,
+    "relocation-model": "static"
 }
--- a/x86_64-bootloader.json
+++ b/x86_64-bootloader.json
@@ -16,7 +16,7 @@
     "os": "none",
     "features": "-mmx,-sse,+soft-float",
     "disable-redzone": true,
-    "panic": "abort",
+    "panic-strategy": "abort",
     "executables": true,
     "relocation_model": "static"
 }
--- a/src/e820.s
+++ b/src/e820.s
@@ -1,5 +1,4 @@
 .section .boot, "awx"
-.intel_syntax noprefix
 .code16
 
 # From http://wiki.osdev.org/Detecting_Memory_(x86)#Getting_an_E820_Memory_Map
--- a/src/stage_1.s
+++ b/src/stage_1.s
@@ -1,6 +1,5 @@
 .section .boot-first-stage, "awx"
 .global _start
-.intel_syntax noprefix
 .code16
 
 # This stage initializes the stack, enables the A20 line, loads the rest of
--- a/src/stage_2.s
+++ b/src/stage_2.s
@@ -1,5 +1,4 @@
 .section .boot, "awx"
-.intel_syntax noprefix
 .code16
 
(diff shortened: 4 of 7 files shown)


x86_64-bootloader.json
```diff
@@ -16,7 +16,6 @@
     "os": "none",
     "features": "-mmx,-sse,+soft-float",
     "disable-redzone": true,
-    "panic": "abort",
-    "executables": true,
-    "relocation_model": "static"
```
bjorn3, line 21: This may work at the moment, but I think it could break with any new rustc version without `"relocation_model": "static"`. With the default `dynamic` relocation model there has to be some code that initializes the GOT and performs other relocations. That it works right now may be because of the linker just so happening to be able to see that none of the called functions are exported and as such can't be interposed at runtime. This allows the use of pc-relative addressing instead of a PLT trampoline that accesses the GOT. There are no guarantees about which rust functions are exported or not.

vinc: Thanks for the explanation! Would there be a way to keep the `relocation_model` field while suppressing the warnings or should I revert the last commit?

bjorn3: I think it should be `relocation-model` with a `-` instead of `_`.

bjorn3: The main branch had the same problem: https://github.com/rust-osdev/bootloader/pull/186

vinc: Ah yes indeed! https://doc.rust-lang.org/rustc/codegen-options/index.html#relocation-model

josephlr (approved):
I also ran into this bug (as a dependency for the blog_os project), this seems to resolve the issue.

phil-opp:
Thanks a lot and sorry for the delay!

Does this mean that the relocation model option was ignored before? Or was it renamed on the rustc side?

josephlr:
> Thanks a lot and sorry for the delay!
> 
> Does this mean that the relocation model option was ignored before? Or was it renamed on the rustc side?

I think it was being ignored, but that the the default (at least in this case) happened to be `"static"`.

phil-opp:
Hmm, from the rust source code it seems like the general default is pic: <https://github.com/rust-lang/rust/blob/aadd6189ad5c81f50d942c584ed1c1b49892765f/compiler/rustc_target/src/spec/mod.rs#L1363>. I haven't found anything yet that indicates that custom targets use static by default...

josephlr:
> Hmm, from the rust source code it seems like the general default is pic: https://github.com/rust-lang/rust/blob/aadd6189ad5c81f50d942c584ed1c1b49892765f/compiler/rustc_target/src/spec/mod.rs#L1363. I haven't found anything yet that indicates that custom targets use static by default...

Oh ya, you're definitely right, I guess things were working because `pic` code will usually work in a `static` context.

phil-opp:
Yeah, that's what I thought. So this PR (and #186) will effectively change the relocation model from `pic` to `static, which might have some side effects. I'm still fine with merging it, but we should keep this in mind in case we see any new bugs afterwards.

phil-opp:
Published as v0.9.19.

Commeownist:
@phil-opp Are you sure you have published `0.9.19` successfully?  [`crates.io`](https://crates.io/crates/bootloader/versions) reports `0.9.18` as the latest version in the `0.9` branch. Could you please recheck and possibly republish?

phil-opp:
Sorry about that, should be fixed now!

vinc:
The release is now available :tada: Thanks :+1:

PR #186, opened 2021-07-17
Fix relocation-model field name in the target spec json

bjorn3:
cc https://github.com/rust-osdev/bootloader/pull/185

### DIFF
--- a/x86_64-bootloader.json
+++ b/x86_64-bootloader.json
@@ -18,5 +18,5 @@
     "disable-redzone": true,
     "panic-strategy": "abort",
     "executables": true,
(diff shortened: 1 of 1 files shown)


josephlr (approved):
LGTM, but I don't have merge permissions for this repo (I think only @phil-opp does?).

phil-opp:
Thanks and sorry for the delay!

As mentioned in #185, this effectively changes the relocation model from `pic` to `static`, so we should keep our eyes open for possible bugs or other side effects.

phil-opp:
Published as v0.10.7

PR #204, opened 2021-11-24
Use new inline assembly syntax for the 0.9 series

bjorn3:


src/main.rs
```diff
@@ -42,8 +42,9 @@ global_asm!(include_str!("video_mode/vga_320x200.s"));
 global_asm!(include_str!("video_mode/vga_text_80x25.s"));
 
 unsafe fn context_switch(boot_info: VirtAddr, entry_point: VirtAddr, stack_pointer: VirtAddr) -> ! {
-    llvm_asm!("call $1; ${:private}.spin.${:uid}: jmp ${:private}.spin.${:uid}" ::
-         "{rsp}"(stack_pointer), "r"(entry_point), "{rdi}"(boot_info) :: "intel");
+    asm!("movq {1}, %rsp; callq *{0}; 0: jmp 0b",
```
bjorn3, line 45: Unfortunately LLVM parses the `0b` as binary literal when using intel syntax. Using a named label is not allowed either as they can cause conflicts when rustc or LLVM decides to duplicate an inline asm block because it for example inlines it. For this reason I had to switch to AT&T syntax.

src/main.rs
```diff
@@ -88,8 +89,10 @@ extern "C" {
 #[no_mangle]
 pub unsafe extern "C" fn stage_4() -> ! {
     // Set stack segment
-    llvm_asm!("mov bx, 0x0
-          mov ss, bx" ::: "bx" : "intel");
+    asm!("push rbx
+          mov bx, 0x0
+          mov ss, bx
+          pop rbx");
```
bjorn3, line 95: The `rbx` register is reserved by LLVM and thus can't be clobbered.

src/main.rs
phil-opp: Have you tried using 2 or larger as label name? https://github.com/rust-lang/rust/issues/76704#issuecomment-815351643 seems to indicate that only 0 and 1 are problematic.

phil-opp (approved):
Looks good to me, thank you!

phil-opp (approved):
Thanks a lot!

phil-opp:
Published as v0.9.21

SlyMarbo:
Would it be possible to push v0.9.21 to crates.io? Many thanks.

phil-opp:
@SlyMarbo Sorry, I forgot to do the actual `cargo publish`. Thanks for reminding me, done!

SlyMarbo:
Amazing! Thanks for all your hard work. It's really appreciated.

PR #209, opened 2021-12-22
Fix `asm` imports on latest nightly

phil-opp:
The `asm` and `global_asm` macros now live in `core::arch` and are no longer part of the prelude on the latest nightlies.


phil-opp:
Published as v0.10.10

PR #210, opened 2022-01-09
Remove feature flag for 'lang_items', 'asm' and 'global_asm'

abachmann:
The feature `asm` and `global_asm` are stable since `1.59.0` and activated by default. So I just removed the feature flags to reduce the number of compiler warnings.

The feature flag `lang_items` has been removed as well as it was no longer used.

### DIFF
--- a/src/bin/bios.rs
+++ b/src/bin/bios.rs
@@ -1,6 +1,3 @@
-#![feature(lang_items)]
-#![feature(global_asm)]
-#![feature(asm)]
 #![no_std]
 #![no_main]
 
--- a/src/bin/uefi.rs
+++ b/src/bin/uefi.rs
@@ -1,7 +1,6 @@
 #![no_std]
 #![no_main]
 #![feature(abi_efiapi)]
-#![feature(asm)]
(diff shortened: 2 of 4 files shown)


phil-opp (approved):
Thank you!

src/bin/bios.rs
```diff
@@ -1,6 +1,4 @@
 #![feature(lang_items)]
```
bjorn3, line 1: Is this still necessary?

abachmann: All tests passed even without it. I'm just wondering that I did not get a warning beforehand. I updated the commit

bjorn3: Unused feature flags haven't been given a warning ever since the introduction of incremental compilation. Only now stabilized feature flags get a warning. `lang_items` is still unstable, but it seems that the current code doesn't use them anymore.

abachmann: Got it. Thanks for the explanation!

phil-opp: @bjorn3 Good catch!

PR #211, opened 2022-01-09
use set_reg method of CS, DS, ES and SS segment structs

abachmann:
I switched to the `set_reg` method to set the `SegmentSelector` for `CS`, `DS`, `ES` and `SS`, because `set_cs`, `load_ds`, `load_es` and  `load_ss` are deprecated in the `x86_64` crate.


phil-opp (approved):
Thanks!

phil-opp:
Published together with #210 as v0.10.11

PR #213, opened 2022-01-19
Logger: nicer font rendering into framebuffer

phip1611:
This PR removes the ugly 8x8 bitmap font and replaces it by a much nicer alternative: [noto-sans-mono-bitmap](https://crates.io/crates/noto-sans-mono-bitmap). It is similar size efficient (doesn't bloat the binary), no_std compatible, uses zero allocations, and no floating point operations.

## Before:
image

## After:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,11 +65,11 @@ dependencies = [
  "conquer-once",
  "displaydoc",
  "fatfs",
- "font8x8",
  "gpt",
  "json",
  "llvm-tools",
  "log",
+ "noto-sans-mono-bitmap",
  "proc-macro2",
  "quote",
  "rsdp",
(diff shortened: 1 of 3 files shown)


phil-opp:
Looks way better, thanks a lot for the PR and for writing this crate!

We should mention the crate in the Rust OSDev newsletter, it's probably useful to many other people too. Do you have some time to write add a few sentences about it to https://github.com/rust-osdev/homepage/pull/84, or should I try to do it?

phip1611:
Thanks @phil-opp! Regarding the newsletter: Sure, great idea! I can do it.

kennystrawnmusic:
These changes introduce an attribute change that generates warnings on Rust 1.60+. I've submitted a [pull request](https://github.com/rust-osdev/bootloader/pull/215) of my own to address this.

bjorn3:
They don't. The warnings are listed under "Unchanged files with check annotations" indicating that this PR didn't change it. It is the result of rustc nightly stabilizing the feature.

PR #214, opened 2022-01-24
Introduce Proper Formatting (`.editorconfig` and `.rustfmt.toml`)

georglauterbach:
This PR adds the `.editorconfig` and `.rustfmt.toml` file to the root of the repository. With these file, we get

1. good IDE formatting support
2. a unified format for the whole crate
3. proper, official Rust RFC styling support

The formatting options were chosen to fit the current style of the code present in this repository. They can be changed easily anytime in the future.

~~Either with a follow-up commit, or with another PR, we can introduce a GitHub actions workflow that lints the code for proper formatting.~~ This already (somewhat) done as of now.

---

Please let me know if there are more options in `.rustfmt.toml` that should be changed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,7 +17,6 @@ indent_size          = 2
 # -------------------------------------------------
 
 [*.{rs,S,asm,ld}]
-indent_style = tab
 indent_size = 4
 
 [Makefile]
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,46 @@
+# -----------------------------------------------
+# ----  General  --------------------------------
+# ----  https://editorconfig.org  ---------------
+# -----------------------------------------------
+
+root = true
+
+[*]
+end_of_line          = lf
+insert_final_newline = true
+charset              = utf-8
+indent_style         = space
+indent_size          = 2
+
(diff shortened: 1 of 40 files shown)


phil-opp:
Thanks for taking the time to create this PR. However, I don't think that we should merge it:

- We already use rustfmt with its default formatting for this repo and even enforce it on CI. I see no reason to change the code style.
- Regarding the editor config: I'm not a big fan of adding and maintaining editor configurations since there are just too many of them. Especially if the configs are only style-related.
  - We mostly have Rust files in this repo, which are formatted using rustfmt anyway. We don't have any Makefile/sh/bash files, so there's no sense in styling them.


Given the above, I'm going to close this.

bjorn3:
> Regarding the editor config: I'm not a big fan of adding and maintaining editor configurations since there are just too many of them. Especially if the configs are only style-related. 

[`.editorconfig`](https://editorconfig.org/) is a standard supported by a lot of editors and with extension support for many more.

phil-opp:
Ah, I didn't know that. However, I'm still not sure if there's much value in adding it. The settings are not enough to satisfy rustfmt, so users need to run rustfmt anyway before commiting.

PR #215, opened 2022-01-27
`maybe_uninit_extra` is no longer feature-gated in Rust 1.60

kennystrawnmusic:
With the most current nightlies, the compiler throws warnings when attempts to build boot images is made because of this ― so time to PR this into the tree to fix this problem.


kennystrawnmusic:
Closing in favor of #218 which happens to have these changes accidentally duplicated.

PR #218, opened 2022-01-30
Let kernels `use` the `Logger`

kennystrawnmusic:
Since A, the next post in Edition 3 of phil-opp/blog_os is supposed to be all about text output, and B, the bootloader already defines a built-in way to do just that ― time to take the `Logger` out of its `binary` cocoon and let everyone `use` it.


kennystrawnmusic:
To explain this further: `unresolved import` compiler errors are currently thrown when a kernel tries to `use bootloader::binary::logger` because `binary` only works when `features = ["binary"]` is added to Cargo.toml — ah, but that's not something a kernel can do without causing the bootloader to panic because it's hardcoded to require the bootloader's built-in custom targets. Hence why I proposed this PR which does 4 things:

1. Move the crates that the `Logger` depends on (`conquer_once`, `noto_sans_mono_bitmap`, `spinning_top` and of course `log`) out of `binary` and into `default` in the bootloader's Cargo.toml
2. Move `src/binary/logger.rs` up one level — to the crate root
3. Move `pub mod logger` from `src/binary/mod.rs` to the crate root
4. Add a `pub use crate::logger` line to `src/binary/mod.rs` for backward compatibility

This approach would allow kernels to simply reuse the bootloader's logger instead of needing to reinvent the wheel.

kennystrawnmusic:
Made a few more changes to push this `Logger` API into a more robust state. Still have yet to hear from the maintainer(s).

phil-opp:
Thanks for the PR, @kennystrawnmusic! I like the idea to give people a way to quickly add some debug printing to their kernels. However, I don't think that we should export the `Logger`/`LockedLogger` API, mainly because I want to avoid the maintenance overhead. There are already good crates for drawing to framebuffers such as [`embedded-graphics`](https://docs.rs/embedded-graphics/latest/embedded_graphics/), so I see no reason to duplicate that functionality here.

So I propose the following:

- Add an optional `logger` feature to this crate that is disabled by default.
- If enabled, make the `init_logger` function public (and move it from the `binary` module to the root).
- Keep the `logger` module private, even if the `logger` feature is enabled.

This way, we provide an easy way to get some output on the screen when getting started, without needing to provide and maintain a full framebuffer writer implementation. Does this sound good to you?

(I also want to mention that I'm currently working on a rewrite of this crate in the `next` branch, which will split off a `bootloader-api` crate. Kernels will only depend on this api crate then, so I'm not sure if we can/should share a logger implementation between the api crate and the main bootloader crate then.)

src/bin/uefi.rs
```diff
@@ -1,7 +1,8 @@
 #![no_std]
 #![no_main]
 #![feature(abi_efiapi)]
-#![feature(maybe_uninit_extra)]
+#![feature(cfg_version)]
+#![cfg_attr(not(version("1.60")), feature(maybe_uninit_extra))]
```
phil-opp, line 5: I think we can just remove the `maybe_uninit_extra` feature unconditionally. Would you mind opening a separate PR for that?

kennystrawnmusic:
Closing now — going to open a couple of new PRs in the next couple of hours to make the recommended changes.

PR #219, opened 2022-01-30
rework `UsedLevel4Entries`

Freax13:
This pr reworks how `UsedLevel4Entries` is created and used. `UsedLevel4Entries` is used to keep track of address ranges that are already in use and dynamically get unused addresses for unconfigured addresses in the config at runtime.

Previously the only addresses that were marked as used where the addresses of the kernel segments. This pr changes the `UsedLevel4Entries` constructor to also mark all the statically configured addresses in the config as used.

This pr also moves the marking of kernel segment addresses into a separate method. This improves support for position independent executables: Previously we just loaded position independent executables with a fixed offset of `0x400000`, but if we also want to get this offset dynamically at runtime, we have to initialize `UsedLevel4Entries` before we mark the kernel segments as used.

### DIFF
--- a/src/binary/level_4_entries.rs
+++ b/src/binary/level_4_entries.rs
@@ -1,9 +1,15 @@
-use core::convert::TryInto;
+use core::{alloc::Layout, convert::TryInto};
+use usize_conversions::IntoUsize;
 use x86_64::{
-    structures::paging::{Page, PageTableIndex},
-    VirtAddr,
+    structures::paging::{Page, PageTableIndex, Size4KiB},
+    PhysAddr, VirtAddr,
 };
 use xmas_elf::program::ProgramHeader;
+
+use crate::{
+    binary::{MemoryRegion, CONFIG},
+    BootInfo,
+};
 
 /// Keeps track of used entries in a level 4 page table.
 ///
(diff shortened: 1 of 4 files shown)


src/binary/level_4_entries.rs
```diff
@@ -13,35 +19,86 @@ pub struct UsedLevel4Entries {
...
+        let end_page_inclusive = Page::<Size4KiB>::containing_address(end_inclusive);
+
+        for p4_index in u16::from(start_page.p4_index())..=u16::from(end_page_inclusive.p4_index())
+        {
+            self.mark_p4_index_as_used(PageTableIndex::new(p4_index));
+        }
+    }
+
+    fn mark_p4_index_as_used(&mut self, p4_index: PageTableIndex) {
+        self.entry_state[usize::from(p4_index)] = true;
```
phil-opp, line 85: Perhaps we should add an assertion here to prevent reusing the same p4 index?

src/config.rs
```diff
@@ -1,3 +1,5 @@
+const PAGE_SIZE: u64 = 4096;
```
phil-opp, line 1: Looks like there are some unused code warnings about this. It probably needs some cfg attribute.

phil-opp (approved):
I left two minor comments, otherwise this looks good to me. Thanks a lot!

src/binary/level_4_entries.rs
Freax13: The problem with that is that we still want to support mapping multiple statically configured address ranges that share the same p4 index. I think if we want to check if statically configured address ranges overlap, we should do it somewhere else because we already lost too much information about the address here.

Alternatively we could replace `UsedLevel4Entries` entirely by something else that tracks the complete addresses not just the p4 indices. Given that we don't map that many addresses, we could probably just use a `StackVec` or something similar.

src/config.rs
Freax13: Should be fixed now.

src/binary/level_4_entries.rs
phil-opp: Ah true! I agree that checking this somewhere else is probably a better solution. But we can do this in a separate PR.

PR #221, opened 2022-02-04
add ASLR

Freax13:
This pr adds optional ASLR. As it turns out generating entropy to seed a random number generated is quite difficult on bare metal. This pr adds function to generate entropy based on `RDRAND`, `RDTSC` and the values in the PIT channels.

### DIFF
--- a/src/binary/entropy.rs
+++ b/src/binary/entropy.rs
@@ -20,7 +20,7 @@ pub fn build_rng() -> ChaCha20Rng {
     ChaCha20Rng::from_seed(seed)
 }
 
-/// Gather entropy by requesting random numbers with `rdrand` instruction if it's available.
+/// Gather entropy by requesting random numbers with `RDRAND` instruction if it's available.
 ///
 /// This function provides excellent entropy (unless you don't trust the CPU vendors).
 fn rd_rand_entropy() -> [u8; 32] {
@@ -41,7 +41,7 @@ fn rd_rand_entropy() -> [u8; 32] {
     entropy
 }
 
-/// Gather entropy by reading the current time with the `rdtsc` instruction if it's available.
+/// Gather entropy by reading the current time with the `RDTSC` instruction if it's available.
 ///
 /// This function doesn't provide particulary good entropy, but it's better than nothing.
 fn tsc_entropy() -> [u8; 32] {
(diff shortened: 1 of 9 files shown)


src/binary/entropy.rs
```diff
@@ -0,0 +1,87 @@
...
+///
+/// This function provides excellent entropy (unless you don't trust the CPU vendors).
+fn rd_rand_entropy() -> [u8; 32] {
+    let mut entropy = [0; 32];
+
+    // Check if the CPU supports `RDRAND`.
+    if let Some(rd_rand) = RdRand::new() {
+        for i in 0..4 {
+            let value = loop {
+                if let Some(value) = rd_rand.get_u64() {
```
bjorn3, line 33: I think this can loop forever in some cases. Eg if there is a hardware defect being detected by the cpu or maybe if the cpu is very hot. Could you limit the iteration count?

Freax13: Yes, you're completely correct. Conveniently Intel's Software Developer's Manual even provides a solution to deal with that.

phil-opp (approved):
Looks really good, thanks a lot!

src/binary/level_4_entries.rs
```diff
@@ -104,23 +111,61 @@ impl UsedLevel4Entries {
...
+            // Count the entries and randomly choose an index in `[0..count)`.
+            let count = free_entries.clone().count();
+            if count == 0 {
+                panic!("no usable level 4 entries found")
+            }
+            let distribution = Uniform::from(0..count);
+            let idx = distribution.sample(rng);
+
+            // Get the index of the free entry.
+            free_entries.nth(idx).unwrap()
```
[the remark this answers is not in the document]
Freax13: No, I just didn't know about that method. Thanks for the suggestion!

phil-opp:
Thanks!

phil-opp:
Published as v0.10.12.

ethindp:
@Freax13 Would you be willing to help me out with my kernel boot problem? I've updated to the latest version and still no go. I'd be happy to move this discussion to Matrix or Discord, whichever you prefer.

Freax13:
> @Freax13 Would you be willing to help me out with my kernel boot problem? I've updated to the latest version and still no go. I'd be happy to move this discussion to Matrix or Discord, whichever you prefer.

Sure, send me an email with your discord tag and I'll add you.

PR #227, opened 2022-03-01
Remove the `asm` and `global_asm` features, which are now stable

### LINKED ISSUE #226: Unnecessary warnings due to inline assembly stabilized in 1.59
Hello,

I am getting the following warnings originating from bootloader. As the warning says, in the latest versions inline assembly has been stabilized and thus these feature gates are causing unnecessary clutter in the terminal.

```rust
warning: the feature `asm` has been stable since 1.59.0 and no longer requires an attribute to enable
 --> src/main.rs:3:12
  |
3 | #![feature(asm)]
  |            ^^^
  |
  = note: `#[warn(stable_features)]` on by default

warning: the feature `global_asm` has been stable since 1.59.0 and no longer requires an attribute to enable
 --> src/main.rs:2:12
  |
2 | #![feature(global_asm)]
  |    
```

Edit: I kept confusing the names, sorry :/

phil-opp:
Fixes #226


src/main.rs
```diff
@@ -1,6 +1,4 @@
 #![feature(lang_items)]
```
bjorn3, line 1: If you compile with panic=abort then `eh_personality` is not necessary and this feature gate can be removed.

phil-opp: Nice! I think only a stable way to build `core` is missing now, then we could compile this crate on stable Rust :tada:.

phil-opp:
Published as v0.9.22

PR #230, opened 2022-04-06
Update stage_3.s

prinzdezibel:
- Data descriptor is not needed in long mode
- 41th bit of code descriptor is unused in long mode

### DIFF
--- a/src/asm/stage_3.s
+++ b/src/asm/stage_3.s
@@ -154,8 +154,7 @@ zero_idt:
 
 gdt_64:
     .quad 0x0000000000000000          # Null Descriptor - should be present.
-    .quad 0x00209A0000000000          # 64-bit code descriptor (exec/read).
-    .quad 0x0000920000000000          # 64-bit data descriptor (read/write).
+    .quad 0x0020980000000000          # 64-bit code descriptor (exec/read).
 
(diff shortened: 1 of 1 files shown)


phil-opp:
Thanks for the PR!

> - Data descriptor is not needed in long mode

You're right that it is not needed in this case, but you still need them for system calls. I'm fine with removing it here.

> - 41th bit of code descriptor is unused in long mode

The AMD manual states that this bit is ignored, but I can't find anything like this in the Intel manual. There are some slight differences between the two sometimes, so I think it's better to keep the `readable` bit if there is any doubt.

prinzdezibel:
@phil-opp , regarding the WRITABLE descriptor flag (41th bit) I am refering to 

https://github.com/rust-osdev/x86_64/blob/8922d8d0b8532324237c2384bdddc91c335bc54f/src/structures/gdt.rs#L208

which states it's not needed in long mode..  Also I've read it your blog_os tutorial, IIRC.

phil-opp:
Looks like this PR fell off my raday, sorry! With https://github.com/rust-osdev/bootloader/pull/232 merged, the `stage_3.s` assembly file no longer exists, so i think we can close this PR.

PR #232, opened 2022-04-10
`v0.11`: Redesign config and build system, split into sub-crates, and provide disk image builder

phil-opp:
This pull requests implements the next version of the `bootloader` crate: `v0.11`. It contains multiple breaking changes:

- **Separate API crate:** The bootloader is now split into two parts: An API crate to make kernels loadable by the bootloader and the actual bootloader implementation. This makes the build process for kernels much easier and faster.
- **New config system:** Instead of configuring the bootloader via a special table in the `Cargo.toml`, the configuration now happens through a normal Rust struct, which is part of the `entry_point!` macro. The macro then serializes the config struct at compile time and places it in a special ELF output section. The compile time serialization happens through a manually implemented `const fn` of the config struct.
- **Load the kernel at runtime:** The bootloader is now able to load files from FAT partitions at runtime. Thus, we don't need to link the kernel into the bootloader executable anymore. As a result, we don't need to recompile the bootloader on kernel changes anymore. We also load the config at runtime from the kernel's ELF section, which eliminates the second reason for recompiling the bootloader as well.
- **Split into sub-crates:** Since the bootloader build process does not need access to the kernel executable or its `Cargo.toml` anymore, we can build the different parts of the bootloader independently. For example, the BIOS boot sector is now a separate crate, and the UEFI bootloader is too.
- **Library to create disk images:** To create an abstraction the complex build steps of the different bootloader executables, we compile them inside cargo build scripts. At the top level, we provide a `bootloader` _library_ crate, which compiles everything as part of its build script. This library includes functions for creating BIOS and UEFI disk images for a given kernel. These functions can be used e.g. from a builder crate or a build script of the downstream operating system.

Taken together, these changes should make the `bootloader` crate much easier to use and also more accessible to contributors.

---

This PR is still a **work in progress**. There are still many things missing, including:

- [x] The BIOS implementation
  - [x] Integrate into build system
  - [x] Pure-Rust boot sector (16-bit real mode) that loads the second stage from a ~FAT partition~ MBR partition entry
  - [x] Pure-Rust second stage (16-bit unreal mode) that initializes the CPU and loads the kernel and third stage from the FAT partition
    - [x] Parse the FAT partition
    - [x] Locate the `kernel-x86_64` file in the root folder
    - [x] Set up protected or unreal mode
    - [x] Load kernel to protected mode memory
    - [x] Create memory map and pass it to stage 3
    - [x] Set up framebuffer
      - [x] search VESA modes for a suitable one
      - [x] fill in framebuffer info in boot information
      - [x] enable the chosen VESA mode
      - [ ] consider config of kernel ELF (maybe later, not clear how we can read the config yet) -> postponed, tracked it https://github.com/rust-osdev/bootloader/issues/281
    - [x] Load third stage and jump to it
  - [x] Stage 3 (protected mode)
    - [x] Use framebuffer logger instead of VGA text mode
    - [x] Set up long mode and identity mapping
      - [x] Identity mapping
      - [x] Compatibility mode
      - [x] 64-bit mode
    - [x] Jump to stage 4
  - [x] Stage 4 (long mode)
    - [x] link with the `common` crate to handle the config and map the kernel
- [x] Adjust README for new version
- [x] Create migration guides
  - [x] From version `0.9.x`
  - [x] From version `0.10.x`
- [x] Clean up the repo and remove all legacy files
- [ ]  For the kernel-requested physical memory mapping we don't want to skip regions. However, instead of mapping the full `0..max_phys_addr` range, we could instead iterate over the reported regions and map only them. See #259 for details. -> we will fix this in a follow-up PR
- [x] Fix BIOS boot error on real machine caused by VESA: https://github.com/rust-osdev/bootloader/pull/232#issuecomment-1243004273
- [x] Improve error message when kernel has no `bootloader-config` section
- [x] Try to find a way to keep the `bootloader-config` section when lto is enabled (see https://github.com/rust-osdev/bootloader/pull/232#issuecomment-1244696577)


src/lib.rs
```diff
@@ -65,65 +65,137 @@ The bootloader can be configured through a `[package.metadata.bootloader]` table
...
+    fs::{self, File},
+    io::{self, Seek},
+    path::Path,
+};
+
+pub fn create_uefi_disk_image(
+    kernel_binary: &Path,
+    out_fat_path: &Path,
+    out_gpt_path: &Path,
+) -> anyhow::Result<()> {
```
Freax13, line 81: This reminds me a lot of the [`bootimage`](https://github.com/rust-osdev/bootimage) crate used in  version 0.9 of the bootloader. Are there any advantages to providing this in a library instead of a binary?

phil-opp: Yeah, I think there are. The problem with `bootimage` was that you needed to install a separate dependency for building the project. The boot system of `bootloader` v0.10 fixed that, but still required you to run a specific binary of the bootloader dependency for building it (which is quite hacky and cumbersome).

With this library-based approach, things should become much simpler and more configurable at the same time. The idea is that you create a top level `os` crate with a `kernel` sub-crate (as a cargo workspace). The `os` crate then has an artifact dependency on the compiled kernel, and a normal dependency on `bootloader`:

```toml
[build-dependencies]
kernel = { path = "kernel", artifact = "bin", target = "x86_64-unknown-none" }
bootloader = "0.11.0-alpha.0"
```

This allows us to create a `build.rs` script that has access to both the compiled kernel and bootloader. All it needs to do is to call the `create_uefi_disk_image` function to combine them:

```rust
fn main() {
    let kernel_path = {
        let path = std::env::var("CARGO_BIN_FILE_KERNEL").expect("kernel binary not found");
        PathBuf::from(path)
    };
    let out_fat_path = kernel_path.with_extension("fat");
    let out_gpt_path = kernel_path.with_extension("gpt");

    bootloader::create_uefi_disk_image(&kernel_path, &out_fat_path, &out_gpt_path)
        .expect("failed to build UEFI disk image");

    println!("cargo:rustc-env=UEFI_FAT_PATH={}", out_fat_path.display());
    println!("cargo:rustc-env=UEFI_GPT_PATH={}", out_gpt_path.display());
}
```

This way, the full disk image is built using just `cargo build`, without requiring any external builder binaries. The `main.rs` of the top-level `os` crate then has access to the path of the bootable disk image via the env variables, so it just needs to e.g. launch QEMU with it. Since you write the QEMU command yourself, you can easily pass whatever arguments you like. Or, you could use a different virtual machine. Or copy it to some specific place, e.g. an USB stick. 

I also plan to make the interface of the `bootloader` crate more fine-grained. The idea is that you could e.g. just create the FAT partition if you like or even just use the bootloader executable and do the whole image creation manually. We should also support to place additional files on the FAT partition, for example compiled userspace applications (which can be another artifact dependency of the `os` crate).

---

You can see an example of the build system in the `tests` directory. The nice thing about using cargo's OUT_DIRs for everything is that concurrent tests are unproblematic since each boot image is placed in a different directory. Tests are also much faster than before because the bootloader does not need to be recompiled for each test anymore.

Freax13: Thanks for the very thorough answer! This all sounds very promising!

Freax13:
I've been working on a PXE bootloader based on the this branch. @phil-opp Would you prefer for me to create the pull request before or after you've finished working on this branch?

phil-opp:
@Freax13 Sounds great!

> Would you prefer for me to create the pull request before or after you've finished working on this branch?

Both are fine, so it depends on the state of your code. If it's ready already, feel free to send a PR! I just don't want to block the v0.11 release any longer than necessary.

x86-16bit-second-stage.json
```diff
@@ -0,0 +1,20 @@
...
+	"max-atomic-width": 64,
+	"position-independent-executables": false,
+	"disable-redzone": true,
+	"target-c-int-width": "32",
+	"target-pointer-width": "32",
+	"target-endian": "little",
+	"panic-strategy": "abort",
+	"os": "none",
+	"vendor": "unknown",
+	"relocation-model": "pie"
```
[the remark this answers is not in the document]
bjorn3: > My understanding was that the resulting binaries for a such target spec will still have relocations that need to be applied before it can be executed.

AFAIK PIE will only include relocations when depending on dynamic libraries.

> When specifying a custom target, we could just switch the relocation model to static, however we obviously can't do that if we want to use x86_64-unknown-none target.

Using the static relocation model would prevent KASLR as it requires a fixed load location at link time.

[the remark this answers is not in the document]
Freax13: > AFAIK PIE will only include relocations when depending on dynamic libraries.

No, PIEs will often still use relocations even when not depending on dynamic libraries, presumably for things where rip relative addressing is not an option e.g. vtables.

> Using the static relocation model would prevent KASLR as it requires a fixed load location at link time.

I was talking about the bootloader, not the kernel.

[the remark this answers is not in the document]
bjorn3: > No, PIEs will often still use relocations even when not depending on dynamic libraries, presumably for things where rip relative addressing is not an option e.g. vtables.

Right, didn't think of that.

> I was talking about the bootloader, not the kernel.

I see. For bios it is fine to use a static relocation model then.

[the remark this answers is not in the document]
Freax13: > I see. For bios it is fine to use a static relocation model then.

`x86_64-unknown-none` would still be nice to have because it's now a Tier 2 target, so it doesn't need `-Zbuild-std` and is therefore compatible with bindeps.

Cargo.toml
```diff
@@ -1,83 +1,49 @@
...
-llvm-tools-build = { version = "0.1", optional = true, package = "llvm-tools" }
-toml = { version = "0.5.1", optional = true }
-serde = { version = "1.0", features = ["derive"], optional = true}
-quote = { version = "1.0", optional = true}
-proc-macro2 = { version = "1.0", optional = true}
+[dependencies.bootloader-x86_64-bios]
+version = "0.1.0-alpha.0"
+path = "bios"
+artifact = ["bin"]
+target = "x86_64-unknown-none"
```
Freax13, line 39: This is blocked by https://github.com/rust-lang/cargo/issues/10405

uefi/src/memory_descriptor.rs
```diff
@@ -0,0 +1,42 @@
...
+            | MemoryType::BOOT_SERVICES_CODE
+            | MemoryType::BOOT_SERVICES_DATA
+            | MemoryType::RUNTIME_SERVICES_CODE
+            | MemoryType::RUNTIME_SERVICES_DATA => {
+                // we don't need this data anymore
+                self.0.ty = MemoryType::CONVENTIONAL;
+            }
+            _ => {}
+        }
+    }
```
Freax13, line 41: This doesn't quite work because we also allocate the kernel's memory with `MemoryType::LOADER_DATA`:
https://github.com/rust-osdev/bootloader/blob/785e82a119d359edd93ef7fc9fc72a7e00f7c9da/uefi/src/main.rs#L201-L208

According to the UEFI spec the memory types `0x80000000..0xFFFFFFFF` are reserved for OS loaders. We could use of those for the kernel allocation.

kennystrawnmusic:
So the BIOS implementation seems to need a lot of work done on it, but what about the UEFI implementation in this redesign? Less to do there?

phil-opp:
@kennystrawnmusic The UEFI implementation is already done. It's much easier to load the kernel and the memory map there.

uefi/src/memory_descriptor.rs
phil-opp: Good catch, thanks! Fixed in 73ae143c4395776ce32def52ba94d87a7671b967

bios/stage-4/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+    // create a new page table hierarchy for the kernel
+    let (kernel_page_table, kernel_level_4_frame) = {
+        // get an unused frame for new level 4 page table
+        let frame: PhysFrame = frame_allocator.allocate_frame().expect("no unused frames");
+        log::info!("New page table at: {:#?}", &frame);
+        // get the corresponding virtual address
+        let addr = phys_offset + frame.start_address().as_u64();
+        // initialize a new page table
+        let ptr = addr.as_mut_ptr();
+        unsafe { *ptr = PageTable::new() };
```
Freax13, line 161: We might want to use `core::ptr::write` here instead. I guess that depends on whether we regard the memory that was just allocated as uninitialized. I'd argue that it is uninitialized, though I don't think this will cause miscompilations considering that this is somewhat magic memory from outside the AM.

bios/stage-4/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+
+#[panic_handler]
+#[cfg(not(test))]
+fn panic(info: &core::panic::PanicInfo) -> ! {
+    unsafe {
+        bootloader_x86_64_common::logger::LOGGER
+            .get()
+            .map(|l| l.force_unlock())
+    };
+    log::error!("{}", info);
```
Freax13, line 218:
```suggestion
    log::error!("{info}");
```

Freax13 (commented):
Initial round of review, mostly nits. This pr is huge, I haven't gotten reviewing all of it yet.

Freax13:
Now that the bootloader is independent from the kernel, we could compile the bootloader in CI and publish it that way. This would also make it easier for people who don't want to use `bootloader::create_boot_partition`, `bootloader::create_uefi_disk_image` etc and want to use their own build scripts instead.

api/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    ($path:path) => {
+        $crate::entry_point!($path, config = &$crate::BootloaderConfig::new_default());
+    };
+    ($path:path, config = $config:expr) => {
+        #[link_section = ".bootloader-config"]
+        pub static __BOOTLOADER_CONFIG: [u8; $crate::BootloaderConfig::SERIALIZED_LEN] = {
+            // validate the type
+            let config: &$crate::BootloaderConfig = $config;
+            config.serialize()
+        };
```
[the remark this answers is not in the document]
phil-opp: Good question, maybe `#[repr(C)]` or `#[repr(C, packed)]` would work too. The potential issue is that the C format depends on the target that the code is compiled for, which might lead to problems if the kernel target and the bootloader target differ too much. For example, an `usize` field would have a different expected size when read from one of the BIOS real mode stages.

Another advantage of using serialization is that we can support normal Rust enums such as Option in the config struct, which are not `repr(C)`.

bios/boot_sector/src/dap.rs
```diff
@@ -0,0 +1,55 @@
...
+    zero: u8,
+    /// Number of sectors to transfer
+    number_of_sectors: u16,
+    /// Offset to memory buffer
+    offset: u16,
+    /// Segment of memory buffer
+    segment: u16,
+    /// Start logical block address
+    start_lba: u64,
+}
```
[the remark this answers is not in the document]
phil-opp: I think that I tried with just `repr(C)` before, but it caused some additional alignment before `start_lba`. But maybe I'm not remembering this correctly.

It definitely should be `#[repr(C, packed)]` though, I'll fix that.

bios/boot_sector/src/dap.rs
```diff
@@ -0,0 +1,55 @@
...
+                "int 0x13",
+                "jc fail",
+                "pop si", // remove error code again
+                "mov si, {1:x}",
+                in(reg) self_addr,
+                out(reg) _,
+                in("ax") 0x4200u16,
+                in("dx") disk_number,
+            );
+        }
```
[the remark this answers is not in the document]
phil-opp: Good idea, done in d8943dd.

bios/stage-2/src/disk.rs
```diff
@@ -0,0 +1,103 @@
...
+        static mut TMP_BUF: AlignedArrayBuffer<1024> = AlignedArrayBuffer {
+            buffer: [0; 512 * 2],
+        };
+        let buf = unsafe { &mut TMP_BUF };
+        assert!(current_sector_offset + len <= buf.buffer.len());
+
+        self.read_exact_into(buf.buffer.len(), buf);
+
+        &buf.buffer[current_sector_offset..][..len]
+    }
```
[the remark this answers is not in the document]
phil-opp: I _think_ we bound the lifetime of the returned `&[u8]` slice to the `&mut self` borrow. So calling `read_exact` should not be possible when the prior slice is still used. Or am I missing something?

We should make it unsafe to construct `DiskAccess` though, to ensure that only a single instance exists.

bios/stage-2/src/memory_map.rs
```diff
@@ -0,0 +1,73 @@
...
+        region_type: 0,
+        acpi_extended_attributes: 0,
+    }; 100],
+);
+
+/// use the INT 0x15, eax= 0xE820 BIOS function to get a memory map
+pub fn query_memory_map() -> Result<&'static mut [E820MemoryRegion], ()> {
+    const SMAP: u32 = 0x534D4150;
+
+    let memory_map = unsafe { MEMORY_MAP.get_mut() };
```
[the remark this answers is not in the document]
phil-opp: Agreed, fixed in fdc630f.

bios/stage-2/src/protected_mode.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // unset protected mode bit again
+    write_cr0(cr0);
+
+    unsafe {
+        asm!("mov ds, {0:x}", in(reg) ds, options(nostack, preserves_flags));
+        asm!("sti");
+
+        asm!("mov bx, 0x0f01", "mov eax, 0xb8000", "mov [eax], bx");
```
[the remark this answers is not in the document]
phil-opp: Good catch, removed in 2eed431.

bios/stage-2/src/disk.rs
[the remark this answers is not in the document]
Freax13: > I _think_ we bound the lifetime of the returned `&[u8]` slice to the `&mut self` borrow. So calling `read_exact` should not be possible when the prior slice is still used. Or am I missing something?

Yes, but there could be multiple `DiskAccess` instances.

> We should make it unsafe to construct `DiskAccess` though, to ensure that only a single instance exists.

That would work, but keep in mind that`DiskAccess` implements `Clone` and IIRC that implementation is also used somewhere to duplicate it.

bios/stage-2/src/protected_mode.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+#[no_mangle]
+pub unsafe fn copy_to_protected_mode(target: *mut u8, bytes: &[u8]) {
+    for (offset, byte) in bytes.iter().enumerate() {
+        let dst = target.wrapping_add(offset);
+        unsafe { asm!("nop", "nop", "mov [{}], {}", in(reg) dst, in(reg_byte) *byte) };
```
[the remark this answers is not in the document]
phil-opp: I think the `nop nop` was just for debugging. I tried to use a normal pointer-based read, but the compiler truncated the pointer to real mode addresses. I removed the nop instructions and added a comment in 2215d00.

bios/stage-2/src/protected_mode.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    GDT.clear_interrupts_and_load();
+
+    // set protected mode bit
+    let cr0 = set_protected_mode_bit();
+
+    // load GDT
+    unsafe {
+        asm!("mov bx, 0x10", "mov ds, bx");
```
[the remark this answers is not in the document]
phil-opp: Good catch, thanks! Fixed in 39b5f16.

bios/stage-2/src/screen.rs
```diff
@@ -0,0 +1,42 @@
+use core::{arch::asm, fmt::Write};
+
+pub fn print_char(c: u8) {
+    let ax = u16::from(c) | 0x0e00;
+    unsafe {
+        asm!("push bx", "mov bx, 0", "int 0x10", "pop bx", in("ax") ax);
```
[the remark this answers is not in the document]
phil-opp: This gives an error: "cannot use register `bx`: rbx is used internally by LLVM and cannot be used as an operand for inline asm"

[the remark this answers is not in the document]
Freax13: That error shouldn't occur when compiling for the bootsector target, `bx`/`rbx` are only used internally by LLVM in 32-bit and 64-bit targets.

bios/stage-3/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+            // jump to 4th stage
+            "pop rax",
+            "pop rdi",
+            "call rax",
+
+            // enter endless loop in case 4th stage returns
+            "2:",
+            "jmp 2b",
+            out("eax") _
+        );
```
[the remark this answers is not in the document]
phil-opp: I'm getting the following error after applying this suggestion: "error: asm outputs are not allowed with the `noreturn` option". Does that mean we can remove the clobbers when using `noreturn`?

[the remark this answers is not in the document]
Freax13: Yes

bios/stage-2/src/disk.rs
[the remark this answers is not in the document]
phil-opp: Ah yeah, you're right. I made it unsafe in e62a933.

bios/stage-2/src/screen.rs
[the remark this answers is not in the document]
phil-opp: Ah, that could be.. Then it's probably just rust-analyzer who is complaining because it runs `cargo check` on the whole workspace. However, given that the current code works too, I'm inclined to keep it that way because it means one warning less in vscode (even if it's a false positive).

bios/stage-4/src/main.rs
phil-opp: Changed in https://github.com/rust-osdev/bootloader/pull/232/commits/26a6aee6ef5311b7dd4882eb2ae007f1c2a7a62c

bios/stage-2/src/screen.rs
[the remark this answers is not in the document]
Freax13: Sure that's fine with me

bios/stage-2/src/protected_mode.rs
```diff
@@ -0,0 +1,155 @@
...
+
+            // jump to third stage
+            "pop eax",
+            "call eax",
+
+            // enter endless loop in case third stage returns
+            "2:",
+            "jmp 2b",
+            out("eax") _
+        );
```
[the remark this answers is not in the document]
phil-opp: Apparently we cannot use clobbers or compiler-selected output registers anymore when using `noreturn`.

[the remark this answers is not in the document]
Freax13: It's the other way around: We don't _have_ to use clobbers. Clobbers tell to compiler that the values of these registers will be different after the asm block, but with `noreturn` there is no code after the asm block, so we don't need to tell the compiler and can just omit all clobbers.

bios/stage-3/src/paging.rs
```diff
@@ -0,0 +1,58 @@
...
+    let l2s = unsafe { LEVEL_2.get_mut() };
+    let common_flags = 0b11;
+    l4.entries[0] = (l3 as *mut PageTable as u64) | common_flags;
+    for i in 0..l2s.len() {
+        let l2 = &mut l2s[i];
+        l3.entries[i] = (l2 as *mut PageTable as u64) | common_flags;
+        let offset = u64::try_from(i).unwrap() * 1024 * 1024 * 1024;
+        for j in 0..512 {
+            l2.entries[j] =
+                (offset + u64::try_from(j).unwrap() * (2 * 1024 * 1024)) | common_flags | (1 << 7);
```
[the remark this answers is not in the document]
phil-opp: Done in 915c25b

bios/stage-2/src/protected_mode.rs
```diff
@@ -0,0 +1,155 @@
...
+        let dst = target.wrapping_add(offset);
+        unsafe { asm!("nop", "nop", "mov [{}], {}", in(reg) dst, in(reg_byte) *byte) };
+        assert_eq!(read_from_protected_mode(dst), *byte);
+    }
+}
+
+#[no_mangle]
+pub unsafe fn read_from_protected_mode(ptr: *mut u8) -> u8 {
+    let res;
+    unsafe { asm!("nop", "nop", "mov {}, [{}]", out(reg_byte) res, in(reg) ptr) };
```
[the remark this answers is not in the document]
phil-opp: Done in 915c25b

bios/stage-2/src/protected_mode.rs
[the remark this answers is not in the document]
phil-opp: Done in 915c25b

phil-opp:
Thanks a lot for taking the time to review! I know that this PR is huge and probably a nightmare to review. Fortunately, the control flow is quite linear, so we mostly need to check whether it is able to load different kernels and whether the given boot information is correct. Maybe it's a good idea to pre-publish an alpha release for this PR and try it out on different projects that are currently using bootloader v0.9 and v0.10?

> Now that the bootloader is independent from the kernel, we could compile the bootloader in CI and publish it that way. This would also make it easier for people who don't want to use bootloader::create_boot_partition, bootloader::create_uefi_disk_image etc and want to use their own build scripts instead. This might be out of the scope for this pr though.

Interesting idea! I agree that there is definitely some value in creating GitHub releases with precompiled artifacts. However, the current form of the BIOS bootloader might make that difficult since it consists of many indepentent executables that have strict requirements about the partition and file system layout. So it's probably better to do this in a follow-up PR and maybe improve the BIOS setup beforehand.

phil-opp:
The CI errors should be fixed by https://github.com/rust-lang/rust/pull/101413, which should be part of the next Rust nightly.

Freax13:
> Thanks a lot for taking the time to review! I know that this PR is huge and probably a nightmare to review. Fortunately, the control flow is quite linear, so we mostly need to check whether it is able to load different kernels and whether the given boot information is correct. 

I'll try to do another round of review today.

> Maybe it's a good idea to pre-publish an alpha release for this PR and try it out on different projects that are currently using bootloader v0.9 and v0.10?

Github codesearch can be used to find projects using the bootloader: ["bootloader = " ("0.9" OR "0.10") path:Cargo.toml](https://cs.github.com/?scopeName=All+repos&scope=&q=%22bootloader+%3D+%22+%28%220.9%22+OR+%220.10%22%29+path%3ACargo.toml)
I've been using a fork of this branch for my own project for a while now (the fork was needed because the bootloader still relied on bindeps and that doesn't work with packages published to crates.io or used with git). I'll switch back to this branch and test.

Freax13:
I'll also test on some real platforms.

src/fat.rs
```diff
@@ -0,0 +1,82 @@
...
+    // choose a file system label
+    let mut label = *b"MY_RUST_OS!";
+    if let Some(path) = files.get(KERNEL_FILE_NAME) {
+        if let Some(name) = path.file_stem() {
+            let converted = name.to_string_lossy();
+            let name = converted.as_bytes();
+            let mut new_label = [0u8; 11];
+            let name = &name[..new_label.len()];
+            let slice = &mut new_label[..name.len()];
+            slice.copy_from_slice(name);
```
Freax13, line 41: This panics if the name is shorter than 11 bytes.

phil-opp:
> Github codesearch can be used to find projects using the bootloader: ["bootloader = " ("0.9" OR "0.10") path:Cargo.toml](https://cs.github.com/?scopeName=All+repos&scope=&q=%22bootloader+%3D+%22+%28%220.9%22+OR+%220.10%22%29+path%3ACargo.toml)

Good idea!



> I've been using a fork of this branch for my own project for a while now (the fork was needed because the bootloader still relied on bindeps and that doesn't work with packages published to crates.io or used with git). I'll switch back to this branch and test.

Ah, I wasn't aware of that. I think the only remaining uses of bindeps are for the test framework, so it should no longer be a problem. But if you encounter any issues, please let me know!

> I'll also test on some real platforms.

That would be great, I haven't done that at all yet.

src/fat.rs
phil-opp: Fixed in 960671c

Freax13:
Hmm, I'm getting a crash on one of my NUCs when booting with the UEFI bootloader:
```
!!!! X64 Exception Type - 0E(#PF - Page-Fault)  CPU Apic ID - 00000000 !!!!
ExceptionData - 0000000000000000  I:0 R:0 U:0 W:0 P:0 PK:0 SS:0 SGX:0
RIP  - 00000000340E964F, CS  - 0000000000000038, RFLAGS - 0000000000010287
RAX  - 0000000034107219, RCX - FFFFFFFD00000000, RDX - 00000000417FFB08
RBX  - 0000000000000000, RSP - 00000000340E7360, RBP - 0000000000000FFF
RSI  - 0000000000000000, RDI - 00000000001536D3
R8   - 0000000034107B88, R9  - 00000000417EC000, R10 - 0000000041588018
R11  - 0000000034108B90, R12 - 00000000340E7650, R13 - 00000000340E73F0
R14  - 8000000000000002, R15 - 0000000000001340
DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
GS   - 0000000000000030, SS  - 0000000000000030
CR0  - 0000000080010013, CR2 - FFFFFFE834107219, CR3 - 0000000033A01000
CR4  - 0000000000000668, CR8 - 0000000000000000
DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 0000000000000400
GDTR - 00000000416CCA98 0000000000000047, LDTR - 0000000000000000
IDTR - 0000000036C43018 0000000000000FFF,   TR - 0000000000000000
FXSAVE_STATE - 00000000340E6FC0
!!!! Can't find image information. !!!!
```
I'll investigate what's going on.

src/main_bak.rs
```diff
@@ -155,19 +149,16 @@ fn main() -> anyhow::Result<()> {
...
         let mut cmd = Command::new(env!("CARGO"));
-        cmd.arg("build").arg("--bin").arg("bios");
-        cmd.arg("--profile").arg("release");
-        cmd.arg("-Z").arg("unstable-options");
-        cmd.arg("--target").arg("x86_64-bootloader.json");
-        cmd.arg("--features")
-            .arg(args.features.join(" ") + " bios_bin");
+        cmd.arg("build").arg("-p").arg("bootloader_first_stage");
+        cmd.arg("--profile").arg("first-stage");
+        cmd.arg("--target").arg("bios/first_stage/x86-16bit.json");
```
Freax13, line 161: This file should probably be removed.

Freax13:
The bounds check for the logger look wrong:
https://github.com/rust-osdev/bootloader/blob/960671c3669c7d1be95f1dfb40afdf9bea8172f8/common/src/logger.rs#L103-L110
It's mixing up width and height. I've ran into index out of bounds panics because of this.

Freax13:
> Hmm, I'm getting a crash on one of my NUCs when booting with the UEFI bootloader:
> 
> ```
> !!!! X64 Exception Type - 0E(#PF - Page-Fault)  CPU Apic ID - 00000000 !!!!
> ExceptionData - 0000000000000000  I:0 R:0 U:0 W:0 P:0 PK:0 SS:0 SGX:0
> RIP  - 00000000340E964F, CS  - 0000000000000038, RFLAGS - 0000000000010287
> RAX  - 0000000034107219, RCX - FFFFFFFD00000000, RDX - 00000000417FFB08
> RBX  - 0000000000000000, RSP - 00000000340E7360, RBP - 0000000000000FFF
> RSI  - 0000000000000000, RDI - 00000000001536D3
> R8   - 0000000034107B88, R9  - 00000000417EC000, R10 - 0000000041588018
> R11  - 0000000034108B90, R12 - 00000000340E7650, R13 - 00000000340E73F0
> R14  - 8000000000000002, R15 - 0000000000001340
> DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
> GS   - 0000000000000030, SS  - 0000000000000030
> CR0  - 0000000080010013, CR2 - FFFFFFE834107219, CR3 - 0000000033A01000
> CR4  - 0000000000000668, CR8 - 0000000000000000
> DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
> DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 0000000000000400
> GDTR - 00000000416CCA98 0000000000000047, LDTR - 0000000000000000
> IDTR - 0000000036C43018 0000000000000FFF,   TR - 0000000000000000
> FXSAVE_STATE - 00000000340E6FC0
> !!!! Can't find image information. !!!!
> ```
> 
> I'll investigate what's going on.

I'm pretty sure this crash is caused by the custom UEFI memory type we use for the kernel. The address in CR2 seems to be linearly dependent on the provided memory type. There used to be a bug in edk2 that stopped those custom memory types working, but that bug was fixed more than 10 years ago in https://github.com/tianocore/edk2/commit/10fe0d814add860a1040e648b1f5782c0de350e6. I don't know why this bug would be present on platforms I've acquired less than a year ago. Just to be sure I checked grub and they also seem to be using the standard memory type `LOADER_DATA` (https://github.com/jiazhang0/grub-efi/blob/cb3fa40dae2cd397e63bd0b1195fec085564ed7e/efi/efimm.c#L91-L95). I also found [someone else coming across the same issue](https://stackoverflow.com/questions/67425959/can-bootloaders-use-custom-uefi-memorytypes). Changing the memory type to `LOADER_DATA` fixed the crash.

Freax13:
> Changing the memory type to `LOADER_DATA` fixed the crash.

With that patched I was able to test the UEFI bootloader in 6 platforms (3 Intel, 3 AMD) and it worked on all of them.

src/main_bak.rs
phil-opp: Done, thanks!

Freax13:
> > Changing the memory type to `LOADER_DATA` fixed the crash.
> 
> With that patched I was able to test the UEFI bootloader on 6 platforms (3 Intel, 3 AMD) and it worked on all of them.

I've published the kernel I used for testing under https://github.com/Freax13/test-kernel.

common/src/logger.rs
```diff
@@ -103,12 +103,10 @@ impl Logger {
                 if self.x_pos >= self.width() {
                     self.newline();
                 }
-                const BITMAP_LETTER_WIDTH: usize =
-                    get_bitmap_width(FontWeight::Regular, BitmapHeight::Size14);
-                if self.y_pos >= (self.height() - BITMAP_LETTER_WIDTH) {
+                let bitmap_char = get_bitmap(c, FontWeight::Regular, BitmapHeight::Size14).unwrap();
+                if self.y_pos >= (self.height() - bitmap_char.height()) {
```
Freax13, line 107: We should also check the width.

phil-opp:
Thanks a lot for testing and debugging this!



> It's mixing up width and height. I've ran into index out of bounds panics because of this.

Fixed in https://github.com/rust-osdev/bootloader/pull/232/commits/47a4c85d96e285f8b208142e455a04eb0a61bcf6.

> Changing the memory type to `LOADER_DATA` fixed the crash.

Interesting! So we need to find another way to recognize the bootloader memory region after allocation, right?

common/src/logger.rs
phil-opp: Ah, you mean in the the `if self.x_pos >= self.width()` check?

Freax13: Yes, unless I'm missing something, that should also be required, shouldn't it?

Freax13:
> Interesting! So we need to find another way to recognize the bootloader memory region after allocation, right?

Yes, but do we even need to mark `LOADER_DATA` regions are usable? AFAICT we only use `LOADER_DATA` for the kernel and the memory map and we don't want to mark either of those as usable after exiting boot services. Unless I'm missing something, us marking the memory map as usable is a bug, isn't it?

uefi/src/memory_descriptor.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    fn usable_after_bootloader_exit(&self) -> bool {
+        match self.0.ty {
+            MemoryType::CONVENTIONAL => true,
+            MemoryType::LOADER_CODE
+            | MemoryType::LOADER_DATA
+            | MemoryType::BOOT_SERVICES_CODE
+            | MemoryType::BOOT_SERVICES_DATA
+            | MemoryType::RUNTIME_SERVICES_CODE
+            | MemoryType::RUNTIME_SERVICES_DATA => {
```
Freax13, line 37: According to the UEFI spec, regions of type `RUNTIME_SERVICES_CODE` and `RUNTIME_SERVICES_DATA` should be preserved by the OS loader and the OS.

phil-opp: Also if you don't use any UEFI runtime services anymore?

Freax13: I think so, yes.

> The memory in this range is to be preserved by the UEFI OS loader and OS
in the working and ACPI S1-S3 states. 
>
> Source: UEFI spec 2.10 Table 7.10 Memory Type Usage after ExitBootServices()

common/src/logger.rs
phil-opp: Added in 130d5f0

phil-opp:
> Yes, but do we even need to mark `LOADER_DATA` regions are usable? AFAICT we only use `LOADER_DATA` for the kernel and the memory map and we don't want to mark either of those as usable after exiting boot services. Unless I'm missing something, us marking the memory map as usable is a bug, isn't it?

I think `loader data` is also used for the UEFI application itself, i.e. storing the data sections of the bootloader executable. The memory map is converted later to a common format and copied into a new region. So we don't need the UEFI-specific memory map anymore after switching to the kernel.

Freax13:
Ah that explains it.

I guess we need something similar to https://github.com/rust-osdev/bootloader/blob/130d5f0ac8b3cd340ea25ce3306068db48112f70/common/src/legacy_memory_region.rs#L113-L130 for the kernel. Damn the custom memory type would have been so much cleaner.

uefi/src/memory_descriptor.rs
phil-opp: Ok, I changed the code to preserve these sections in 667e57f.

phil-opp:
> I guess we need something similar to [...]

Done in https://github.com/rust-osdev/bootloader/pull/232/commits/08e4b5829bf5882d9d396e641e32b65de72704b2

Freax13:
https://github.com/rust-osdev/bootloader/blob/667e57f552e214f9c19848306e03b00d91a8114f/api/src/info.rs#L175-L177
This is unsound because `create_buffer` creates a mutable reference which is not allowed to alias other references created by `buffer()`.

bios/stage-3/src/screen.rs
```diff
@@ -0,0 +1,130 @@
...
+                if self.x_pos >= self.width() {
+                    self.newline();
+                }
+                const BITMAP_LETTER_WIDTH: usize =
+                    get_bitmap_width(FontWeight::Regular, BitmapHeight::Size14);
+                if self.y_pos >= (self.height() - BITMAP_LETTER_WIDTH) {
+                    self.clear();
+                }
+                let bitmap_char = get_bitmap(c, FontWeight::Regular, BitmapHeight::Size14).unwrap();
+                self.write_rendered_char(bitmap_char);
```
Freax13, line 86: Same thing as https://github.com/rust-osdev/bootloader/pull/232#discussion_r967805500

uefi/src/main.rs
```diff
@@ -0,0 +1,441 @@
...
+
+static SYSTEM_TABLE: VeryUnsafeCell<Option<SystemTable<Boot>>> = VeryUnsafeCell::new(None);
+
+struct VeryUnsafeCell<T>(UnsafeCell<T>);
+
+impl<T> VeryUnsafeCell<T> {
+    const fn new(v: T) -> Self {
+        Self(UnsafeCell::new(v))
+    }
+}
```
Freax13, line 45: This should probably renamed to `RacyCell` to be consistent with 
https://github.com/rust-osdev/bootloader/blob/7088b5344442aadbc138d247944ca34ffdccf4cc/bios/common/src/racy_cell.rs#L5

Freax13 (commented):
I just finished a review of the entire code, left two small comments that should get addressed, but after that I think it's good to go from my side.

README.md
```diff
@@ -12,60 +12,56 @@ You need a nightly [Rust](https://www.rust-lang.org) compiler with the `llvm-too
...
-- [ ] Transform this "Future Plans" list into issues and a roadmap.
+This project is split into three separate entities:
+
+- A [`bootloader_api`](./api) library with the entry point, configuration, and boot info definitions.
+  - Kernels should include this library as a normal cargo dependency.
+  - The provided `entry_point` macro will encode the configuration settings into a separate ELF section of the compiled kernel executable.
+- [BIOS](./bios) and [UEFI](./uefi) binaries that contain the actual bootloader implementation.
+  - The implementations share a higher-level [common library](./common).
+  - Both implementations load the kernel at runtime from a FAT partition. This FAT partition is created
+  - The configuration is read from a special section of the kernel's ELF file, which is created by the `entry_point` macro of teh `bootloader_api` library.
```
Freax13, line 61:
```suggestion
  - The configuration is read from a special section of the kernel's ELF file, which is created by the `entry_point` macro of the `bootloader_api` library.
```

api/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+        #[export_name = "_start"]
+        pub extern "C" fn __impl_start(boot_info: &'static mut $crate::BootInfo) -> ! {
+            // validate the signature of the program entry point
+            let f: fn(&'static mut $crate::BootInfo) -> ! = $path;
+
+            // ensure that the config is used so that the linker keeps it
+            $crate::__force_use(&__BOOTLOADER_CONFIG);
+
+            f(boot_info)
+        }
```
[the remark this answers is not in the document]
phil-opp: Nice trick, thanks!

Freax13 (approved):

phil-opp:
Thanks a lot a for reviewing!

I try to create add a basic migration guide later, then I think we can merge :tada:.

asensio-project:
Hello,

I tried this version with the test-kernel from @Freax13, with UEFI works fine. But with legacy BIOS on bare metal doesn't work.
I get a PANIC on bios/stage-2/src/main.rs on the line 111:10, the cause is an unwrap on a None value of an Option.

Thanks for this amazing project.

phil-opp:
@asensio-project Thanks for testing! Seems like this is related to the VESA config here:

https://github.com/rust-osdev/bootloader/blob/f95e9a97190c118f94f914b0c2483b137b8be39d/bios/stage-2/src/main.rs#L108-L111

I'll look into it!

TheBotlyNoob:
getting 
```rust
UEFI bootloader started; trying to load kernel
panicked at 'called `Option::unwrap()` on a `None` value', common\src\lib.rs:67:27
```
It seems to be coming from https://github.com/rust-osdev/bootloader/blob/next/common/src/lib.rs#L67-L69 and,
only happens when I compile in release mode with `lto = true`

toothbrush7777777:
> It seems to be coming from https://github.com/rust-osdev/bootloader/blob/next/common/src/lib.rs#L67-L69 and, only happens when I compile in release mode with `lto = true`

@TheBotlyNoob You probably need a custom linker script specifying to KEEP those custom sections.

phil-opp:
@TheBotlyNoob Thanks for reporting! There was a bug in the `entry_point` macro so that the config was not marked as used properly. I pushed a fix in [076bea9](https://github.com/rust-osdev/bootloader/pull/232/commits/076bea979fe148a23eb2eb35469726f4f87dbab5) and added a test to prevent regressions.

phil-opp:
> Seems like this is related to the VESA config here:
> 
> https://github.com/rust-osdev/bootloader/blob/f95e9a97190c118f94f914b0c2483b137b8be39d/bios/stage-2/src/main.rs#L108-L111
>
> I'll look into it!

Whew, that took a lot of debugging. It turns out that there were two bugs in the VESA code related to segmented memory. Apparently QEMU doesn't use segments in that case, so it was only triggered by real hardware. I pushed fixes for the bugs in https://github.com/rust-osdev/bootloader/pull/232/commits/6eb0abbb7e5b4b6088fb49452c7dc976183a946a and https://github.com/rust-osdev/bootloader/pull/232/commits/a6d69e9b4bd9daf4dc006d25b9d11d6c06edaab2. Now it boots fine with both UEFI and BIOS on my old laptop.

Freax13:
Just did another test run: Tested the UEFI bootloader on 6 platforms (4x Intel, 2x AMD) and the BIOS bootloader on 1 platform (Intel). all successful

uefi/src/main.rs
```diff
@@ -0,0 +1,441 @@
...
+        kernel: kernel_page_table,
+        kernel_level_4_frame,
+    }
+}
+
+fn init_logger(st: &SystemTable<Boot>, config: BootloaderConfig) -> (PhysAddr, FrameBufferInfo) {
+    let gop = st
+        .boot_services()
+        .locate_protocol::<GraphicsOutput>()
+        .expect("failed to locate gop");
```
Freax13, line 369: We shouldn't panic when there's not graphics output.

Freax13: Started working on a pr: https://github.com/Freax13/bootloader/tree/optional-framebuffer

It should work, but it needs to be tested some more. I got a little side-tracked reinvestigating https://github.com/rust-osdev/bootloader/pull/232#issuecomment-1242928162

Freax13:
> > Hmm, I'm getting a crash on one of my NUCs when booting with the UEFI bootloader:
> > ```
> > !!!! X64 Exception Type - 0E(#PF - Page-Fault)  CPU Apic ID - 00000000 !!!!
> > ExceptionData - 0000000000000000  I:0 R:0 U:0 W:0 P:0 PK:0 SS:0 SGX:0
> > RIP  - 00000000340E964F, CS  - 0000000000000038, RFLAGS - 0000000000010287
> > RAX  - 0000000034107219, RCX - FFFFFFFD00000000, RDX - 00000000417FFB08
> > RBX  - 0000000000000000, RSP - 00000000340E7360, RBP - 0000000000000FFF
> > RSI  - 0000000000000000, RDI - 00000000001536D3
> > R8   - 0000000034107B88, R9  - 00000000417EC000, R10 - 0000000041588018
> > R11  - 0000000034108B90, R12 - 00000000340E7650, R13 - 00000000340E73F0
> > R14  - 8000000000000002, R15 - 0000000000001340
> > DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
> > GS   - 0000000000000030, SS  - 0000000000000030
> > CR0  - 0000000080010013, CR2 - FFFFFFE834107219, CR3 - 0000000033A01000
> > CR4  - 0000000000000668, CR8 - 0000000000000000
> > DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
> > DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 0000000000000400
> > GDTR - 00000000416CCA98 0000000000000047, LDTR - 0000000000000000
> > IDTR - 0000000036C43018 0000000000000FFF,   TR - 0000000000000000
> > FXSAVE_STATE - 00000000340E6FC0
> > !!!! Can't find image information. !!!!
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll investigate what's going on.
> 
> I'm pretty sure this crash is caused by the custom UEFI memory type we use for the kernel. The address in CR2 seems to be linearly dependent on the provided memory type. There used to be a bug in edk2 that stopped those custom memory types working, but that bug was fixed more than 10 years ago in [tianocore/edk2@10fe0d8](https://github.com/tianocore/edk2/commit/10fe0d814add860a1040e648b1f5782c0de350e6). I don't know why this bug would be present on platforms I've acquired less than a year ago. Just to be sure I checked grub and they also seem to be using the standard memory type `LOADER_DATA` (https://github.com/jiazhang0/grub-efi/blob/cb3fa40dae2cd397e63bd0b1195fec085564ed7e/efi/efimm.c#L91-L95). I also found [someone else coming across the same issue](https://stackoverflow.com/questions/67425959/can-bootloaders-use-custom-uefi-memorytypes). Changing the memory type to `LOADER_DATA` fixed the crash.

Turns out that there is a very similar (yet unpatched) bug that someone else found about three years ago: https://bugzilla.tianocore.org/show_bug.cgi?id=2372 and this is actually what went wrong when I was testing. I confirmed by attaching a debugger and reverse engineering the part of the firmware that crashed. This bug hasn't been patched yet, so, sadly, we still can't use custom memory types.

uefi/src/main.rs
Freax13: I opened a pr to fix this: #268.

Freax13:
> Thanks for looking into this! Even if this is fixed at some point, there will still be systems with older firmware in the wild, right? So we should probably just keep using the default memory type.

Yup, I don't think we'll ever get to the point where all devices are patched. It would be nice not to rely on the workaround, it's really a shame that this bug exists, I liked the custom memory types :disappointed:. I mainly looked into this because I couldn't believe that reputable vendors would push firmware that contained bugs that have been patched more than 10 years ago.

Freax13:
When the alpha was released, docs.rs failed to build documentation for it: https://docs.rs/crate/bootloader/0.11.0-alpha.0/builds/539317. The build.rs script tried to build the uefi and bios bootloaders, but didn't have network access to download the dependencies and failed.

Cargo.toml
```diff
@@ -1,89 +1,46 @@
...
-version = "0.10.13"
+version = "0.11.0-alpha"
 authors = ["Philipp Oppermann <dev@phil-opp.com>"]
 license = "MIT/Apache-2.0"
 description = "An experimental x86_64 bootloader that works on both BIOS and UEFI systems."
 repository = "https://github.com/rust-osdev/bootloader"
-edition = "2018"
-build = "build.rs"
+edition = "2021"
+resolver = "2"
```
Freax13, line 11: `resolver = 2` is the default in Rust 2021.

phil-opp: Ah yes, I'll remove it!

phil-opp: Done as part of https://github.com/rust-osdev/bootloader/pull/232/commits/2ca6ebd4d9a8e8b52fa48eaae8dab476b3bee8c9

phil-opp:
I think this is ready to be merged :tada:.

Issue #87, opened 2019-12-17
PXE support (with iPXE)


GuillaumeDIDIER:
Would it be possible to have support to boot using PXE or iPXE (with iPXE downloading the bootimage.bin using TFTP for instance).

This would require changing the logic to load the files in memory (which iPXE probably takes care of).

This could be poretty helpful for development on real hardware, as it wuld allow to network boot instead of going through the write to USB manipulations.

GuillaumeDIDIER:
gamozolabs/orange_slice may be an intersting ressource given that it has a PXE bootloader.

phil-opp:
This would be definitely useful. However, I don't have the time to look into this right now, as there are other roadmap items with a higher priority (support for multiboot and UEFI).

If someone wants to work on this, I would be happy to provide mentoring for the bootloader integration.

ethindp:
Shouldn't this just work out of the box? Considering that the bootloader is linked with the kernel, no intermediate steps should be necessary other than flashing the iPXE ROM and setting up an iPXE boot script to download the binary.

Freax13:
I've opened a pr that implements PXE support for the UEFI bootloader.

Is there any demand for implementing Legacy PXE support? I have been working on a simple implementation for a Legacy PXE  (see https://github.com/Freax13/bootloader/tree/legacy-pxe). If there's not much demand for it I'm not sure if it's worth integrating: Most newer devices don't implement Legacy PXE anymore, the code's not that pretty (mostly copied from the BIOS bootloader, but different in a few places that make it hard to split out the common parts) and the only platform I have to test behaves wildly different than QEMU.

ChocolateLoverRaj:
Btw if you don't want to deal with TFTP (like me) you can use iPXE with http to load this bootloader. With something like this:
```
#!ipxe
module http://gaming-computer:64017/kernel-x86_64
module http://gaming-computer:64017/ramdisk
chain http://gaming-computer:64017/efi/boot/bootx64.efi
```
but I will probably just use http since iPXE doesn't support modern certificate algorithms.

PR #237, opened 2022-05-10
add UEFI PXE support

Freax13:
This pr implements a UEFI PXE bootloader by extending the existing UEFI bootloader to load a kernel from a TFTP server if the DevicePath of the LoadedImage has a device implementing PXE BaseCode protocol.

~~This pr needs some [unreleased new features](https://github.com/rust-osdev/uefi-rs/pull/417) of the `uefi-rs` crate, so we'll have to wait for a new release before merging this branch (I switched the `uefi-rs` dependency to use the master branch for now). This is the reason I'm marking the pr as a draft.~~

Related to #87

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -123,6 +123,8 @@ pub fn create_bios_disk_image(
     Ok(())
 }
 
+/// Prepare a folder for use with booting over UEFI_PXE. The dhcp server should
+/// have the filename option set to `bootloader`.
 pub fn create_uefi_pxe_tftp_folder(kernel_binary: &Path, out_path: &Path) -> anyhow::Result<()> {
     let bootloader_path = Path::new(env!("UEFI_BOOTLOADER_PATH"));
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -123,8 +123,11 @@ pub fn create_bios_disk_image(
     Ok(())
 }
 
-/// Prepare a folder for use with booting over UEFI_PXE. The dhcp server should
-/// have the filename option set to `bootloader`.
+/// Prepare a folder for use with booting over UEFI_PXE.
+///
+/// This places the bootloader executable under the path "bootloader". The
+/// DHCP server should set the filename option to that path, otherwise the
+/// bootloader won't be found.
 pub fn create_uefi_pxe_tftp_folder(kernel_binary: &Path, out_path: &Path) -> anyhow::Result<()> {
     let bootloader_path = Path::new(env!("UEFI_BOOTLOADER_PATH"));
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,6 +78,7 @@ use std::{
 mod fat;
 mod gpt;
 mod mbr;
+mod pxe;
 
 const KERNEL_FILE_NAME: &str = "kernel-x86_64";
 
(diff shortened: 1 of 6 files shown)


Freax13:
The CI failure seems unrelated to this pr.

phil-opp:
Thanks a lot for the PR, looks good overall! I'm happy to merge it once you think it's ready, just let me know.

Freax13:
> Thanks a lot for the PR, looks good overall! I'm happy to merge it once you think it's ready, just let me know.

I'll add some more comments, but after that it should be done

Freax13:
> Thanks a lot for the PR, looks good overall! I'm happy to merge it once you think it's ready, just let me know.

It's ready now! :tada:

tests/runner/src/lib.rs
```diff
@@ -42,3 +46,40 @@ pub fn run_test_kernel(kernel_binary_path: &str) {
...
+    strip_ansi_escapes::Writer::new(std::io::stderr())
+        .write_all(&child_output.stderr)
+        .unwrap();
+    strip_ansi_escapes::Writer::new(std::io::stderr())
+        .write_all(&child_output.stdout)
+        .unwrap();
+
+    match child_output.status.code() {
+        Some(33) => {}                     // success
+        Some(35) => panic!("Test failed"), // success
```
phil-opp, line 82:
```suggestion
        Some(35) => panic!("Test failed"),
```

src/lib.rs
```diff
@@ -121,3 +122,14 @@ pub fn create_bios_disk_image(
 
     Ok(())
 }
+
+/// Prepare a folder for use with booting over UEFI_PXE. The dhcp server should
+/// have the filename option set to `bootloader`.
```
phil-opp, line 127: > The dhcp server should have the filename option set to `bootloader`.

Could you clarify what this means and why it is necessary?

phil-opp (approved):
Looks good! I left two minor annotations, then we can merge.

src/lib.rs
Freax13: That's part of how PXE works under the hood, the PXE client asks the DHCP server where to boot from. Very quick history lesson:
DHCP is based on BOOTP. BOOTP was designed to allow diskless workstations to boot from the network. The server-sent Offer and ACK packets would have their "Boot file name" and a "Next server IP address" fields populated to point the client to the TFTP server and the bootloader file.
DHCP still has those fields, but nowadays most DHCP servers (e.g. in home network WiFi routers) are mostly used to give out dynamic IP addresses and DNS configuration and don't populate the "Boot file name" field because they don't care about PXE.
The reason I explicitly mention what the bootloader file name option should be set to is that there's no standardized/default value for this field. The correct value depends entirely on where the bootloader file is located on the TFTP server which happens to be in `bootloader` with the current implementation. We also have to do this in the tests: We set the `bootfile` option on the network device which causes QEMU's integrated DHCP server to set the "Boot file name" field in the response packets:
https://github.com/rust-osdev/bootloader/blob/f975de96c65d2c850edeae3e8b852f2cf78c14a6/tests/runner/src/lib.rs#L64-L67

phil-opp: Thanks a lot for the detailed explanation!

PR #240, opened 2022-06-29
implement a new context switch

### LINKED ISSUE #239: don't leave parts of the bootloader in the kernel's address space
While implementing finer granular ASLR I came across this comment:
https://github.com/rust-osdev/bootloader/blob/ac46d0455b41c11e5d316348d068df1c495ce0af/src/binary/level_4_entries.rs#L40
We mark the first 512GiB of the address space as unusable for dynamically generated addresses. I think we do this because we identity map the context switch code into kernel memory and this code most likely resides within the first 512GiB of the address space:
https://github.com/rust-osdev/bootloader/blob/a445433010960ec5d8a8b94a85fcac16a00489b5/src/binary/mod.rs#L166-L181

This causes a number of (admittedly small and unlikely) problems:
- The identity mapped pages could overlap with the kernel or other mappings
- We don't expose the identity mapped addresses to the kernel in `Mappings`
- An attacker could make use of the identity mapped pages to defeat ASLR
- We mark so a lot of usable memory as unusable and because of that we can't check for overlaps because there would be a lot of false positives. We currently just ignore overlaps.

We could probably work around those problems while still mapping parts of the bootloader into the kernel's address space, but I'd like to propose another solution: We use another very short lived page table to do the context switch. This page table would only map a few pages containing code that switches to the kernel's page table. Importantly, we would set the page table up in such a way that the kernel's entrypoint is just after the page table switch instruction, so we don't have to use any code to jump to the kernel, it would simply be the next instruction. 
I don't think we could reliably map such code into the bootloader's address space because we'd have to map the code just before the kernel's entrypoint which could be close to bootloader's code, so that's why I want to use a short-lived page table.

We also identity map a GDT into the kernel's address space:
https://github.com/rust-osdev/bootloader/blob/a445433010960ec5d8a8b94a85fcac16a00489b5/src/binary/mod.rs#L183-L193
We should probably make the GDT's location configurable and expose it in `Mappings`.

I'd be happy to work on a pr for this.

Freax13:
This pr reimplements the context switch in such a way that we don't need to map any of the bootloader's memory into the kernel's address space (see #239 for more info). This pr also makes the GDT base address configurable instead of identity mapping it into the kernel's address space.

Depends on https://github.com/rust-osdev/x86_64/pull/385
Closes #239


phil-opp:
Thanks a lot for implementing this! Unfortunately, I don't have much time right now, but I try my best to review this as soon as possible.

Freax13:
> Thanks a lot for implementing this! Unfortunately, I don't have much time right now, but I try my best to review this as soon as possible.

No worries, take your time!

Freax13:
Just rebased onto the latest `next` branch.

phil-opp:
Thanks a lot! I'm really sorry for the delay in reviewing.

Given that this changes a critical part of the boot process, I think we should get https://github.com/rust-osdev/bootloader/pull/232 merged and released first since many people tested that version already. This change would then become `v0.11.1`, which would make it easier to compare the behavior and detect possible regressions. Would that be ok for you?

Freax13:
Sure, sounds good to me!

Freax13:
Why was this PR closed? Couldn't we just have switched the target branch to `main`?

bjorn3:
Given that the delection of the target branch and the closing of this PR happened at the same time I thinl github auto-closed it.

phil-opp:
Oh wow, looks like a race condition on GitHub's side. Normally, it just changes the base branch in this case (and it did so for https://github.com/rust-osdev/bootloader/pull/282), but apparently it handled the "branch deleted" action first for this PR.

phil-opp:
Ok, restoring the branch allowed me to reopen this PR. I now changed the base branch manually.

Freax13:
Thanks!

PR #243, opened 2022-07-25
mitigate undefined symbol _fltused

tsatke:
This hopefully fixes the build until the issue is resolved in the linker.

Ref rust-lang/rust#62785

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,3 +127,7 @@ macro_rules! entry_point {
         }
     };
 }
+
+#[used]
+#[no_mangle]
+pub static _fltused: i32 = 0;


bjorn3:
The UEFI target is compiled with soft-float enabled. Shouldn't that suppress usage of _fltused? Or am I misunderstanding it's purpose? Also `_fltused` is still emitted on UEFI by compiler-builtins: https://github.com/rust-lang/compiler-builtins/blob/dffad9b98ac33c4e22a08397cfbcc52262ab1676/src/x86_64.rs#L77-L82 That will conflict with this definition, right?

tsatke:
@bjorn3 it fixed all problems with my bootloader fork.
However, the issue seems fixed with today's rust nightly, closing this.

PR #247, opened 2022-07-31
Export the ACPI tables

kennystrawnmusic:
As per [this comment](https://github.com/phil-opp/blog_os/discussions/1014#discussioncomment-3288198) on the heap allocation post in @phil-opp's OS development tutorial (as well as the comment that this is a reply to which has a detail of the problem that this pull request aims to solve), the process of parsing the ACPI tables can be challenging to say the least, especially if things like handlers need to not only be handwritten but also depend on the pre-existence of mapped pages. I have therefore decided to submit my solution mentioned in that comment as this pull request.

Also as stated in that very comment, this code has indeed been tested ― by first writing it as part of following Edition 3's UEFI application tutorial. Most of this code is merely copied-and-pasted from the implementation I wrote as part of the result of following that. If anyone wants to test the changes, uncomment Line 144 of the changed `uefi/src/main.rs` file.


kennystrawnmusic:
So I'm noticing this in the CI:

```
running 3 tests
Unable to init server: Could not connect: Connection refused
gtk initialization failed
Unable to init server: Could not connect: Connection refused
gtk initialization failed
test basic_boot ... FAILED
test check_boot_info ... FAILED
Unable to init server: Could not connect: Connection refused
gtk initialization failed
test should_panic ... FAILED
```

Why would the server be refusing the connection? Doesn't seem right.

uefi/src/main.rs
```diff
@@ -81,6 +130,25 @@ fn main_inner(image: Handle, mut st: SystemTable<Boot>) -> Status {
...
+    if let Some(rsdp) = config
+        .find(|e| matches!(e.guid, ACPI_GUID | ACPI2_GUID))
+        .map(|e| e.address)
+    {
+        match unsafe { AcpiTables::from_rsdp(BootAcpi, rsdp as usize) } {
+            Ok(tables) => {
+                let portable_tables = PortableAcpiTables::new(tables);
+                ACPI_TABLES.get_or_init(move || portable_tables);
+                // Debug
+                // info!("Interrupt model: {:#?}", ACPI_TABLES.get().unwrap().info.interrupt);
```
kennystrawnmusic, line 144: To test these changes, uncomment this line.

uefi/Cargo.toml
```diff
@@ -11,5 +11,10 @@ license = "MIT/Apache-2.0"
 bootloader_api = { version = "0.1.0-alpha.0", path = "../api" }
 bootloader-x86_64-common = { version = "0.1.0-alpha.0", path = "../common" }
 log = "0.4.14"
-uefi = "0.16.0"
+uefi = { version = "0.16.0", features = ["alloc"] }
 x86_64 = "0.14.8"
+acpi = "4.1.0"
+
+[dependencies.conquer-once]
+version = "0.3.2"
```
bjorn3, line 19: nit: missing trailing newline

uefi/src/main.rs
```diff
@@ -32,8 +48,35 @@ use x86_64::{
     PhysAddr, VirtAddr,
 };
 
+pub mod portable_acpi;
+use portable_acpi::PortableAcpiTables;
 mod memory_descriptor;
 
+pub static ACPI_TABLES: OnceCell<PortableAcpiTables> = OnceCell::uninit();
```
bjorn3, line 55: UEFI is single threaded, so we don't need a spinlock. `VeryUnsafeCell` like `SYSTEM_TABLE` would probably work fine.

bjorn3: Also `pub` in an executable doesn't have any effect.

uefi/src/portable_acpi.rs
```diff
@@ -0,0 +1,347 @@
...
+        sdt::Signature,
+        AcpiHandler, AcpiTables, AmlTable,
+        InterruptModel::{self, *},
+        PowerProfile, Sdt,
+    },
+    alloc::vec::Vec,
+};
+
+/// Drop-in replacement for `acpi::platform::interrupt::Apic` that uses slices instead of vectors
+/// for easy transportation across the UEFI Boot Services boundary
```
bjorn3, line 20: Slices don't have a stable ABI either. Also as there is no `#[repr(C)]` the compiler is free to reorder fields.

bjorn3:
I don't get why you need to do this in the bootloader. Other kernels like linux don't do it in the kernel either.

kennystrawnmusic:
> I don't get why you need to do this in the bootloader. Other kernels like linux don't do it in the kernel either.

See [this comment](https://github.com/phil-opp/blog_os/discussions/1014#discussioncomment-3282555) on the heap allocation post. The lack of examples in the `acpi` crate’s documentation has been quite the stumbling block.

bjorn3:
I don't see why it would deadlock in the kernel, jut not the bootloader. I think you just made a mistake somewhere in your kernel. You can probably attach gdb to qemu to see where it hangs, or print to the serial port at various points to see which prints are hit and which are never reached due to the deadlock.

kennystrawnmusic:
> I don't see why it would deadlock in the kernel, jut not the bootloader. I think you just made a mistake somewhere in your kernel. You can probably attach gdb to qemu to see where it hangs, or print to the serial port at various points to see which prints are hit and which are never reached due to the deadlock.

I tried the second approach you mentioned — added `log::info!(“This is a test”)` both before and after the line where the ACPI tables are initiated. It printed before I called `AcpiTables::from_rsdp` using the address passed by the bootloader but not after.

Might be improper page alignment in which case a review would be necessary; not entirely sure though.

kennystrawnmusic:
Going to close this for now; not exactly the easiest thing to refactor.

PR #256, opened 2022-09-03
[Logger] Don't panic if unsupported char

asensio-project:
Hello,

Currently, when we print a char in the screen if it is unsupported on noto-sans-mono-bitmap it panics.
Now the noto-sans-mono-bitmap supports the '■' char, we can use it instead of panicking.

Changes included in this PR:
    - Update noto-sans-mono-bitmap (Cargo.toml)
    - Don't panic if an unsupported char is found (src/binary/logger.rs)

Thanks!

### DIFF
--- a/src/binary/logger.rs
+++ b/src/binary/logger.rs
@@ -108,7 +108,10 @@ impl Logger {
                 if self.y_pos >= (self.height() - BITMAP_LETTER_WIDTH) {
                     self.clear();
                 }
-                let bitmap_char = get_bitmap(c, FontWeight::Regular, BitmapHeight::Size14).unwrap();
+                let bitmap_char = get_bitmap(c, FontWeight::Regular, BitmapHeight::Size14)
+                    .unwrap_or(
+                        get_bitmap('■', FontWeight::Regular, BitmapHeight::Size14).unwrap()
+                    );
                 self.write_rendered_char(bitmap_char);
             }
         }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ rand = { version = "0.8.4", optional = true, default-features = false }
 rand_chacha = { version = "0.3.1", optional = true, default-features = false }
 
 [dependencies.noto-sans-mono-bitmap]
-version = "0.1.2"
+version = "0.1.6"
 default-features = false
 features = ["regular", "size_14"]
 optional = true


src/binary/logger.rs
```diff
@@ -108,7 +108,10 @@ impl Logger {
                 if self.y_pos >= (self.height() - BITMAP_LETTER_WIDTH) {
                     self.clear();
                 }
-                let bitmap_char = get_bitmap(c, FontWeight::Regular, BitmapHeight::Size14).unwrap();
+                let bitmap_char = get_bitmap(c, FontWeight::Regular, BitmapHeight::Size14)
+                    .unwrap_or(
+                        get_bitmap('■', FontWeight::Regular, BitmapHeight::Size14).unwrap()
```
bjorn3, line 113: Maybe � (Replacement Character)?

asensio-project: What are you referring? � it's not in the noto-sans-mono-bitmap crate yet.

bjorn3: > it's not in the noto-sans-mono-bitmap crate yet.

Really? I would have expected it to be. Guess it is fine to leave it like this.

asensio-project: OK, later we can ask to @phip1611 for adding that char. Two days ago the other char was not.
See this [issue](https://github.com/phip1611/noto-sans-mono-bitmap-rs/issues/4) for more information.

phip1611: > Really? I would have expected it to be. Guess it is fine to leave it like this.

Maintainer here: I'm about to add unicode ranges as selectable cargo features. Thus, it will be available soon, hopefully.

asensio-project: tho... What?

phip1611: > tho... What?

https://www.urbandictionary.com/define.php?term=tho :)

asensio-project: Sorry, I never listened that. I'm from Spain and English in Spain is quite bad.
Thanks, now I have one word more.

asensio-project: So we merge it, or we wait for the other char which is more descriptive?

phip1611: Update regarding unicode features in the `noto-sans-mono-bitmap`. I experimented with it but even adding a "reasonable amount" of unicode ranges results in a 12 GB large crate, as all the characters must be pre-rasterized and shipped with the crate. 

This is unacceptable. I decided that `noto-sans-mono-bitmap` has to strive for minimality. If you want the whole unicode range @asensio-project , switch to user-mode and use the floating point unit. You will be able to use dynamic rendering of all the chars you can think of.

I'll drop the whole approach but update the crate documentation

asensio-project: OK, I will create a PR for adding the � char to noto-sans-mono-bitmap-rs and then we can merge it for the main branch and the next branch of the bootloader. @phip1611

phip1611:
I opened https://github.com/rust-osdev/bootloader/pull/267 to bump the `noto-sans-mono-bitmap` to `0.2.0` (unreleased yet). This will make this PR obsolete.

asensio-project:
Closing this in favor of #267.

Issue #259, opened 2022-09-20
better handling of very large memory maps


hawkw:
Currently, the bootloader has some issues handling BIOS memory maps that contain very high addresses, such as around the 1TB mark, and/or very large regions (in excess of 500 GB). Depending on the `bootloader` version and configuration, memory maps with high addresses or large regions may result in crashes or degrade boot performance significantly.

In particular, the following issues have been observed:

- v0.9.x of `bootloader` contains an assertion that may be triggered when the memory map contains an address too big for a single PML4 entry if the `map_physical_memory` feature flag is enabled: https://github.com/rust-osdev/bootloader/blob/42da77b47558478c5a948d27cac80767b484eacc/src/main.rs#L289-L291 (see also phil-opp/blog-os#1138)
- v0.10.x of `bootloader` doesn't contain that assertion, but does have offset selection code that assumes no region will require multiple PML4 entries: https://github.com/rust-osdev/bootloader/blob/ac46d0455b41c11e5d316348d068df1c495ce0af/src/binary/level_4_entries.rs#L172-L174
- v0.10.x of `bootloader` exhibits very long boot times when a memory map contains a reserved region at a ~1TB offset. This is because the bootloader will identity map all pages up to the highest reserved address in the memory map using 4K pages, and does this twice (once for the bootloader itself, and a second time when setting up physical memory mappings for the kernel). This results in a very slow boot process.

These issues are relevant because AMD systems with an IOMMU have a reserved hole close to the 1TB mark. In recent QEMU versions (>= v7.1.0), QEMU will report this region as reserved in its `e380` BIOS memory map, resulting in assertion failures or boot performance degradation. I would assume this would cause issues when booting on real AMD hardware, as well.

Some changes that would improve how `bootloader` handles high addresses in the memory map (many of which were suggested by @phil-opp on Gitter):

- [ ] Change the automatic offset selection to support regions that need multiple entries in the level 4 page table (v0.9.x, v0.10.x, and probably v0.11.x)
- [ ] Don't identity map reserved regions at all when performing the initial identity mapping for the bootloader itself (v0.10.x and probably v0.11.x)
  - The framebuffer would still need to be identity mapped so that the bootloader can write to it.
- [ ]  Use 2MB rather than 4KB pages when identity mapping the kernel address space, which would improve performance (v0.10.x and v0.11.x)
[...]

PR #260, opened 2022-09-22
Limit BIOS bootloader's `max_phys_addr` to 4 GiB

hawkw:
When determining the maximum physical address for the BIOS bootloader's identity mapping, we currently use the highest address in the E380 memory map, no matter how high it is. However, the bootloader runs in protected mode, and therefore, it cannot address more than 4 GiB of memory. We can save some time by not identity mapping addresses over 4 GiB, as the bootloader cannot address them anyway.

This commit changes the BIOS bootloader to skip addresses over 4 GiB when determining the maximum physical address.

This is one of the changes described in issue #259.

### DIFF
--- a/src/bin/bios.rs
+++ b/src/bin/bios.rs
@@ -78,6 +78,7 @@ fn bootloader_main(
     use bootloader::binary::{
         bios::memory_descriptor::E820MemoryRegion, legacy_memory_region::LegacyFrameAllocator,
     };
+    const GIGABYTE: u64 = 4096 * 512 * 512;
 
     let e820_memory_map = {
         let ptr = usize_from(memory_map_addr.as_u64()) as *const E820MemoryRegion;
@@ -86,6 +87,10 @@ fn bootloader_main(
     let max_phys_addr = e820_memory_map
         .iter()
         .map(|r| r.start_addr + r.len)
+        // Don't consider addresses > 4GiB when determining the maximum
+        // physical address for the bootloader, as we are in  protected mode and
+        // cannot address more than 4 GiB of memory anyway.
+        .filter(|&addr| addr < GIGABYTE * 4)
         .max()
         .expect("no physical memory regions found");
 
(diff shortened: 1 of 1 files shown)


src/bin/bios.rs
```diff
@@ -86,6 +87,10 @@ fn bootloader_main(
     let max_phys_addr = e820_memory_map
         .iter()
         .map(|r| r.start_addr + r.len)
+        // Don't consider addresses > 4GiB when determining the maximum
+        // physical address for the bootloader, as we are in  protected mode and
+        // cannot address more than 4 GiB of memory anyway.
+        .filter(|&addr| addr < GIGABYTE * 4)
```
phil-opp, line 93: It might be better to cap the address at 4 GiB instead of filtering the region out completely. For example, we don't want to filter a hypothetical region that goes from 2GiB to 5GiB.

For example: 

```suggestion
        .map(|&addr| u64::min(addr, GIGABYTE * 4))
```

phil-opp (commented):
Thank you!

src/bin/bios.rs
bjorn3: Or alternatively move the `u64::min` call around the whole expession calculating `max_phys_addr`.

hawkw: I wasn't sure whether we would prefer to potentially map only part of a region if it crosses the 4GiB boundary, or skip mapping that region entirely. I can change that!

hawkw: Done in 4da46e1a72b7279241eabf3f25abfb1f10fc69d7

phil-opp (approved):

phil-opp:
Interesting, thanks for testing! I suspect that the performance difference is caused by the TLB flushes. For the identity mapping we flush each entry after creation. For the physical memory mapping, on the other hand, we don't do any flushing because the bootloader never accesses the memory. The TLB will be flushed completely when switching to the kernel address space anyway.

So I think we should be able to improve the performance further by doing a full TLB flush after creating the mappings instead of flushing each new entry separately.

hawkw:
@phil-opp good idea --- should i make that change in this PR or on a separate branch?

phil-opp:
However you prefer :).

hawkw:
Sounds good, I'll make that change separately, then!

phil-opp:
Published together with https://github.com/rust-osdev/bootloader/pull/265 and https://github.com/rust-osdev/bootloader/pull/263 as v0.10.13

PR #261, opened 2022-09-22
Include `llvm-tools-preview` in `rust-toolchain.toml`

hawkw:
Building the bootloader test and example kernels requires the `llvm-tools-preview` Rust toolchain component to be installed. Rustup can automatically install additional toolchain components if the `rust-toolchain` file includes a `components` section.

This commit adds a configuration in the `rust-toolchain` file to tell Rustup to install `llvm-tools-preview` automatically. This requires switching the toolchain file from the legacy format to [the TOML format]. Therefore, I've also renamed the file from `rust-toolchain` to `rust-toolchain.toml`, so that it gets highlighted as a TOML file by most editors by default.

[the TOML format]: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

### DIFF
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,1 +1,0 @@
-nightly
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+[toolchain]
+channel = "nightly"
+components = [
+    "llvm-tools-preview",
+]


Freax13:
This will run into https://github.com/actions-rs/toolchain/issues/126

phil-opp:
> This will run into [actions-rs/toolchain#126](https://github.com/actions-rs/toolchain/issues/126)

With all components specified in the `rust-toolchain.toml` we should not need the toolchain action anymore. The GitHub action runners include `rustup` by default, which will automatically install the specified toolchain and components. The `next` branch already includes this change, maybe it's possible to cherry-pick.

hawkw:
@phil-opp I can cherry-pick the changes to the CI configuration, one sec.

phil-opp:
This was implemented as part of the `v0.11` rewrite. Thanks for proposing this!

PR #263, opened 2022-09-22
fix `get_free_address` for large sizes (0.10)

### LINKED ISSUE #262: panic on large memory regions when `map-physical-memory` is enabled
Trying to test out a higher half kernel using PR #229, the bootloader panics when the `map-physical-memory`, `map-framebuffer`, and `dynamic-range-start` configurations are all present.

I have the following configurations in my `Cargo.toml`:
```toml
[package.metadata.bootloader]
map-physical-memory = true
# the kernel is mapped into the higher half of the virtual address space.
dynamic-range-start = "0xFFFF_8000_0000_0000"

[patch.crates-io]
# patch `bootloader` to use https://github.com/rust-osdev/bootloader/pull/229
# for dynamic range address configuration support.
#
# remove this patch once rust-osdev/bootloader#229 makes it into a release
# version.
bootloader = { git = "https://github.com/rust-osdev/bootloader", rev = "ac46d0455b41c11e5d316348d068df1c495ce0af" }
```

I get the following panic:
https://github.com/rust-osdev/bootloader/blob/ac46d0455b41c11e5d316348d068df1c495ce0af/src/binary/mod.rs#L355
which fails on a `ParentEntryHugeFrame` error:
image

Freax13:
This pr fixes `get_free_address` for large sizes (more than 2^39).

Part of #259
Closes #262

### DIFF
--- a/src/binary/level_4_entries.rs
+++ b/src/binary/level_4_entries.rs
@@ -132,7 +132,7 @@ impl UsedLevel4Entries {
     }
 
     /// Returns the first index of a `num` contiguous unused level 4 entries and marks them as
-    /// used. If `CONFIG.aslr` is enabled, this will return random available entries.
+    /// used. If `CONFIG.aslr` is enabled, this will return random contiguous available entries.
     ///
     /// Since this method marks each returned index as used, it can be used multiple times
     /// to determine multiple unused virtual memory regions.
--- a/src/binary/level_4_entries.rs
+++ b/src/binary/level_4_entries.rs
@@ -131,19 +131,18 @@ impl UsedLevel4Entries {
         }
     }
 
-    /// Returns an unused level 4 entry and marks it as used. If `CONFIG.aslr` is
-    /// enabled, this will return a random available entry.
+    /// Returns the first index of a `num` contiguous unused level 4 entries and marks them as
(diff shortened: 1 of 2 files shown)


Freax13:
The mac os CI job timed out, but the other jobs are fine. The timeout is caused by #259.

src/binary/level_4_entries.rs
```diff
@@ -131,19 +131,18 @@ impl UsedLevel4Entries {
         }
     }
 
-    /// Returns an unused level 4 entry and marks it as used. If `CONFIG.aslr` is
-    /// enabled, this will return a random available entry.
+    /// Returns the first index of a `num` contiguous unused level 4 entries and marks them as
+    /// used. If `CONFIG.aslr` is enabled, this will return random available entries.
```
hawkw, line 135: It might be worth explicitly stating
```suggestion
    /// used. If `CONFIG.aslr` is enabled, this will return random contiguous available entries.
```
for clarity?

hawkw (approved):

src/binary/level_4_entries.rs
Freax13: Agreed, thanks!

phil-opp (approved):
Looks good, thank you!

PR #265, opened 2022-09-23
Only perform a single TLB flush after identity mapping

hawkw:
This commit changes the BIOS bootloader to only flush the TLB once, after it has identity-mapped every physical memory frame. This should be a bit more efficient, as we don't perform a separate `invlpg` for every page table entry we create, and instead only flush the entire thing by reloading `CR3` when we're actually ready to use it.

This is based on a suggestion by @phil-opp, here:
https://github.com/rust-osdev/bootloader/pull/260#issuecomment-1255418145

This change doesn't actually seem to make all that big an impact in boot times on QEMU v7.1 on its own, relative to PR #260, but it might make an additional improvement on top of that PR.


phil-opp (approved):
Looks good, thanks!

PR #267, opened 2022-10-01
fb/logger: bump noto-sans-mono-bitmap + improvements

phip1611:
Hi,

### Update
This is now ready for merge.

----
### Original Post

I'm the author of the [noto-sans-mono-bitmap](https://github.com/phip1611/noto-sans-mono-bitmap-rs/)-crate and about to release `0.2.0` with breaking changes and many improvements ([CHANGELOG](https://github.com/phip1611/noto-sans-mono-bitmap-rs/blob/main/CHANGELOG.md)).

As the bootloader crate is the only/the biggest user, I'd like to leave room for API discussions before I release the crate.

This MR bumps the `noto-sans-mono-bitmap` to 0.2.0 (unreleased yet) and performs several other small improvements to the framebuffer logger.

Any comments? If there are no comments, I'll release the crate in 2-4 days and switch this MR from draft state to ready

What's new?
- more unicode ranges and they are selectable as cargo features 
- the internal structure was refactored by a lot which enables crate users to better specify which features they want and which they do not want
- the API changed along with the documentation to better fit the use case

**Example**
Please note the `�` fallback character in the `BIOS boot` logging message, which is also new.
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,17 +56,23 @@ rand = { version = "0.8.4", optional = true, default-features = false }
 rand_chacha = { version = "0.3.1", optional = true, default-features = false }
 
 [dependencies.noto-sans-mono-bitmap]
-version = "0.1.2"
+version = "0.2.0"
 default-features = false
-features = ["regular", "size_14"]
+features = [
+    "regular",
+    "size_16",
+    "unicode-basic-latin",
+    # required for the fallback char '�'
+    "unicode-specials"
+]
 optional = true
 
 [build-dependencies]
 llvm-tools-build = { version = "0.1", optional = true, package = "llvm-tools" }
 toml = { version = "0.5.1", optional = true }
-serde = { version = "1.0", features = ["derive"], optional = true}
-quote = { version = "1.0", optional = true}
-proc-macro2 = { version = "1.0", optional = true}
+serde = { version = "1.0", features = ["derive"], optional = true }
+quote = { version = "1.0", optional = true }
+proc-macro2 = { version = "1.0", optional = true }
 
 [features]
 default = []
@@ -94,6 +100,6 @@ default-target = "x86_64-unknown-linux-gnu"
 [package.metadata.release]
 dev-version = false
 pre-release-replacements = [
-    { file="Changelog.md", search="# Unreleased", replace="# Unreleased\n\n# {{version}} – {{date}}", exactly=1 },
+    { file = "Changelog.md", search = "# Unreleased", replace = "# Unreleased\n\n# {{version}} – {{date}}", exactly = 1 },
 ]
 pre-release-commit-message = "Release version {{version}}"
--- a/src/binary/logger.rs
+++ b/src/binary/logger.rs
@@ -1,10 +1,13 @@
+use crate::binary::logger::font_constants::BACKUP_CHAR;
 use crate::boot_info::{FrameBufferInfo, PixelFormat};
 use conquer_once::spin::OnceCell;
 use core::{
     fmt::{self, Write},
     ptr,
 };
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -56,17 +56,24 @@ rand = { version = "0.8.4", optional = true, default-features = false }
...
-version = "0.1.2"
+path = "../noto-sans-mono-bitmap"
 default-features = false
-features = ["regular", "size_14"]
+features = [
+    "regular",
+    "size_16",
+    "unicode-basic-latin",
+    "unicode-latin-1-supplement",
+    "unicode-latin-extended-a",
```
phip1611, line 66: is this needed?

Cargo.toml
```diff
@@ -56,17 +56,24 @@ rand = { version = "0.8.4", optional = true, default-features = false }
 rand_chacha = { version = "0.3.1", optional = true, default-features = false }
 
 [dependencies.noto-sans-mono-bitmap]
-version = "0.1.2"
```
phip1611, line 59: just a draft MR; will change once I released `0.2.0`

asensio-project:
Looks great!

src/binary/logger.rs
```diff
@@ -39,8 +75,7 @@ impl log::Log for LockedLogger {
 
     fn log(&self, record: &log::Record) {
         let mut logger = self.0.lock();
-        writeln!(logger, "{}:    {}", record.level(), record.args()).unwrap();
-        logger.add_vspace(LOG_SPACING);
```
phil-opp, line 43: Is there any specific reason for removing this? This gave us a bit of extra padding around individual log messages (which might be multi-line).

phil-opp (commented):
Looks good to me, I like the new crate features!

Cargo.toml
phil-opp: I think we only need basic latin. All the log messages and file names shoud be ASCII.

src/binary/logger.rs
phip1611: it still exists, but it is now located here: https://github.com/rust-osdev/bootloader/pull/267/files#diff-516264e516b0fbbd881351d365dbf0af59b2a151f200ce3fdc8abd4f0b99cbf7R106

I think, it makes more sense there.

phil-opp: There is a slight difference for multi-line messages. Before, the padding was only added between "paragraphs", now it is also added for every line break in long messages. But either way is fine with me, so we can keep it like this.

Cargo.toml
phip1611: alright

phip1611:
> Looks good to me, I like the new crate features!

this is now ready to merge!

Cargo.toml
```diff
@@ -94,6 +100,6 @@ default-target = "x86_64-unknown-linux-gnu"
 [package.metadata.release]
 dev-version = false
 pre-release-replacements = [
-    { file="Changelog.md", search="# Unreleased", replace="# Unreleased\n\n# {{version}} – {{date}}", exactly=1 },
```
phip1611, line 97: yikes, my IDE did this automatically

Cargo.toml
```diff
@@ -56,17 +56,23 @@ rand = { version = "0.8.4", optional = true, default-features = false }
...
+    "unicode-basic-latin",
+    # required for the fallback char '�'
+    "unicode-specials"
+]
 optional = true
 
 [build-dependencies]
 llvm-tools-build = { version = "0.1", optional = true, package = "llvm-tools" }
 toml = { version = "0.5.1", optional = true }
-serde = { version = "1.0", features = ["derive"], optional = true}
```
phip1611, line 67: yikes, my IDE did this automatically

phil-opp (approved):
Great, thanks!,

PR #268, opened 2022-10-06
allow booting without a UEFI graphics output

Freax13:
This pr makes booting without a UEFI graphics output possible.


phil-opp (approved):
Looks good, thanks!

PR #269, opened 2022-10-06
implement read-only relocations

Freax13:
This pr implements GNU_RELRO program header handling to make relocations read-only. This is a common exploit mitigation. The `x86_64-unknown-none` target already has [the relro level set to full](https://github.com/rust-lang/rust/blob/607b8296e07cc1bf5b95eeb60a21b5af684f7631/compiler/rustc_target/src/spec/x86_64_unknown_none.rs#L17), so this should increase security without any actions needed by the user.

### DIFF
--- a/src/binary/load_kernel.rs
+++ b/src/binary/load_kernel.rs
@@ -127,6 +127,14 @@ where
             }
         }
 
+        // Mark some memory regions as read-only after relocations have been
+        // applied.
+        for program_header in self.elf_file.program_iter() {
+            if let Type::GnuRelro = program_header.get_type()? {
+                self.inner.handle_relro_segment(program_header);
+            }
+        }
+
         self.inner.remove_copied_flags(&self.elf_file).unwrap();
 
         Ok(tls_template)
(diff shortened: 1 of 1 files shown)


src/binary/load_kernel.rs
```diff
@@ -496,6 +504,44 @@ where
...
+    /// This is a security mitigation used to protect memory regions that
+    /// need to be writable while applying relocations, but should never be
+    /// written to after relocations have been applied.
+    fn handle_relro_segment(&mut self, program_header: ProgramHeader) {
+        let start = self.virtual_address_offset + program_header.virtual_addr();
+        let end = start + program_header.mem_size();
+        let start = VirtAddr::new(start);
+        let end = VirtAddr::new(end);
+        let start_page = Page::containing_address(start);
+        let end_page = Page::containing_address(end - 1u64);
```
phil-opp, line 519: It is possible that the segment shares a page with other segments that should stay writable? In that case, marking the whole page as read-only would lead to issues...

Freax13: I don't think so, but I could find any spec that says so.
The spec only says:
> The array element specifies the location and size of a segment which may be made read-only after relocation shave been processed.

(https://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic.html)

I found different implementations:
glibc's runtime linker seems to round down the start as well as the end (https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/dl-reloc.c;h=756bf950f6f98aa04e4ee23d202f27c26775584e;hb=refs/heads/master#l353) whereas android's runtime linker seems to round the start down and the end up (https://android.googlesource.com/platform/bionic/+/master/linker/linker_phdr.cpp#888). 

LLVM's lld always makes sure that the end of the segment is always page aligned (https://github.com/llvm/llvm-project/blob/c3e073bcbdc523b0f758d44a89a6333e38bff863/lld/ELF/Writer.cpp#L2633-L2638), but doesn't say anything about the start.

I think the current implementation is fine.

phil-opp: Ok, thanks for checking!

phil-opp (approved):
Thanks!

PR #283, opened 2022-11-13
Prepare for v0.11 release and fix remaining issues

phil-opp:
Prepares the new `v0.11` merged in #232 for publishing.

TODO:

- [x] update changelog
- [x] create git tags after publishing

Fixes #291 
Fixes #286

PR #289, opened 2022-11-28
support higher half position independent kernels

### LINKED ISSUE #285: Support position independent code in higher half
When a PIC kernel is located in higher half, we currently try to allocate 33554177 entries in the level 4 page table, which of course fails. This is the reason that we currently use the `static` relocation model for our higher half test:

https://github.com/rust-osdev/bootloader/blob/4e2b8d7e4b4fad3c7aefd7f8174f86ec2119ecc7/Cargo.toml#L87

I think the issue is in these lines:

https://github.com/rust-osdev/bootloader/blob/4e2b8d7e4b4fad3c7aefd7f8174f86ec2119ecc7/common/src/load_kernel.rs#L64-L75

Looks like we only consider the end address for the size calculation, without taking the start address into account. I think we should be able to fix this by using the difference between the max and min address as size instead.

Freax13:
This pr adds support for loading position independent kernels with a base address other than 0.

Currently the base address is ignored. In the future, we could try to map the kernel at it's base address if it's non zero. Doing this right now would be difficult because we don't have accurate book keeping on what addresses have already been allocated (See #239). I propose we take another look at this once finer granular ASLR is implemented (finer granular ASLR requires more accurate book keeping).

Closes #285

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ rustflags = [
     "-C",
     "link-args=--image-base 0xFFFF800000000000",
     "-C",
-    "relocation-model=static",                   # pic in higher half not supported yet
+    "relocation-model=pic",
     "-C",
     "code-model=large",
 ]
--- a/build.rs
+++ b/build.rs
@@ -52,6 +52,7 @@ fn build_uefi_bootloader(out_dir: &Path) -> PathBuf {
         // local build
         cmd.arg("--path").arg("uefi");
         println!("cargo:rerun-if-changed=uefi");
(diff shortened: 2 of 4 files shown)


Freax13:
It happened again, GitHub automatically closed this pr when #283 was merged.

phil-opp:
Sorry about that! Maybe the automatic PR base change does not work across forks?

For now I disabled the automatic deletion of merged branches for this repo to avoid this issue in the future.

phil-opp (approved):
Sorry for the late review. Thanks a lot for this PR, the code is really nice!

Issue #288, opened 2022-11-27
Add a QUIET feature


asensio-project:
Hello, 

I think that would be good idea adding a cargo feature for avoiding the bootloader to print messages in the screen until loading the kernel.
This way we don't need to clear the screen before printing messages.

Thanks.

Freax13:
Note that you might have to clear the screen anyways because some BIOS' will show the manufacturers logo on startup.

asensio-project:
A published a new PR #303.
It's ready for review.

PR #292, opened 2022-11-30
QUIET Feature

asensio-project:
Hello,

This PR adds a Quiet feature for avoiding the bootloader to write to the screen.
It sets the log filter in error when the feature is enabled.

Thanks.


asensio-project:
Closes #288

phil-opp:
Thanks for submitting this PR! Unfortunately, I don't think that it works this way because the `common` library is built as part of the build script, which doesn't pass any cargo features. I think it would make more sense to add this as a runtime config option in the [`BootloaderConfig`](https://docs.rs/bootloader_api/0.11.0/bootloader_api/config/struct.BootloaderConfig.html) struct.

asensio-project:
> Thanks for submitting this PR! Unfortunately, I don't think that it works this way because the `common` library is built as part of the build script, which doesn't pass any cargo features. I think it would make more sense to add this as a runtime config option in the [`BootloaderConfig`](https://docs.rs/bootloader_api/0.11.0/bootloader_api/config/struct.BootloaderConfig.html) struct.

Thanks for the feedback, I published a PR, #303 for applying the changes that you requested. It's ready for review.

PR #300, opened 2022-12-22
Fix Create Disk Image Example

AlexJMohr:
The example code didn't compile. This patch fixes that.

Let me know if it needs any changes, and thanks for making this awesome crate!

### DIFF
--- a/docs/create-disk-image.md
+++ b/docs/create-disk-image.md
@@ -29,7 +29,7 @@ test-kernel = { path = "kernel", artifact = "bin", target = "x86_64-unknown-none
 
 [dependencies]
 # used for UEFI booting in QEMU
-ovmf_prebuilt = "0.1.0-alpha.1"
+ovmf-prebuilt = "0.1.0-alpha.1"
 
 [workspace]
 members = ["kernel"]
--- a/docs/create-disk-image.md
+++ b/docs/create-disk-image.md
@@ -38,14 +38,14 @@ members = ["kernel"]
 ```rust
 // build.rs
 
-use std::path::Path;
+use std::path::PathBuf;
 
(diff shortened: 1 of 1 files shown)


phil-opp (approved):
Thanks a lot! Looks like I forgot to try this snippet.

One minor nit: I think the `Path::new` call is simple enough to do apply it in the `out_dir`/`kernel` bindings directly instead of introducing new `out_dir_path`/`kernel_path` bindings. I left some suggestions below for this.

AlexJMohr:
@phil-opp I agree, that looks simpler.

asensio-project:
Hello, I think that in the Cargo.toml file you should change ``ovmf_prebuilt`` to ``ovmf-prebuilt`` because the first package doesn't exists in crates.io but the second does.

phil-opp:
> The compiler complains about `temporary value is freed at the end of this statement`

I think we can use `PathBuf` instead of `Path` to take ownership of the string. I pushed https://github.com/rust-osdev/bootloader/pull/300/commits/63284c610bca46b7f021d842e2300570598c3f99 with that change.

> Hello, I think that in the Cargo.toml file you should change `ovmf_prebuilt` to `ovmf-prebuilt` because the first package doesn't exists in crates.io but the second does.

Thanks, done in https://github.com/rust-osdev/bootloader/pull/300/commits/0d24cef05198572efa3eff9dd2cd80561f883a7d.

phil-opp:
Could you try whether the new version works for you? Then we can finally merge this!

asensio-project:
The new version is working very good, thanks!

phil-opp:
Ok great, thanks for testing!

PR #301, opened 2022-12-30
Add support for adding a ramdisk to the disk image

jasoncouture:
This adds support for adding a ramdisk file to the disk image.

I will follow up in a future PR to add support for loading and passing this to the kernel in a future PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,11 +37,9 @@ impl BiosBoot {
     }
 
     /// Add a ramdisk file to the image
-    pub fn with_ramdisk(&self, ramdisk_path: &Path) -> Self {
-        Self {
-            kernel: self.kernel.to_owned(),
-            ramdisk: Some(ramdisk_path.to_owned()),
-        }
+    pub fn set_ramdisk(&mut self, ramdisk_path: &Path) -> &mut Self {
+        self.ramdisk = Some(ramdisk_path.to_owned());
+        self
     }
 
     /// Create a bootable UEFI disk image at the given path.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,12 +17,14 @@ mod mbr;
 mod pxe;
 
 const KERNEL_FILE_NAME: &str = "kernel-x86_64";
+const RAMDISK_FILE_NAME: &str = "ramdisk-x86_64";
 const BIOS_STAGE_3: &str = "boot-stage-3";
 const BIOS_STAGE_4: &str = "boot-stage-4";
 
 /// Create disk images for booting on legacy BIOS systems.
 pub struct BiosBoot {
     kernel: PathBuf,
+    ramdisk: Option<PathBuf>,
 }
 
 impl BiosBoot {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -61,12 +72,18 @@ impl BiosBoot {
...
 
         let mut files = BTreeMap::new();
-        files.insert(KERNEL_FILE_NAME, self.kernel.as_path());
+        files.insert(KERNEL_FILE_NAME, kernel_path);
         files.insert(BIOS_STAGE_3, stage_3_path);
         files.insert(BIOS_STAGE_4, stage_4_path);
-
+        if has_rd_path {
+            files.insert(RAMDISK_FILE_NAME, ramdisk_path);
+        }
```
phil-opp, line 86: This seems like a good use case for `if let`:

```suggestion
        if let Some(ramdisk_path) = &self.ramdisk {
            files.insert(RAMDISK_FILE_NAME, ramdisk_path);
        }
```

This way, we don't need the additional bindings in lines 75-78.

src/lib.rs
```diff
@@ -121,10 +148,17 @@ impl UefiBoot {
...
+        let ramdisk_path = binding.unwrap_or(Path::new("no-such-file"));
+        let kernel_path = self.kernel.as_path();
         let mut files = BTreeMap::new();
         files.insert("efi/boot/bootx64.efi", bootloader_path);
-        files.insert(KERNEL_FILE_NAME, self.kernel.as_path());
+        files.insert(KERNEL_FILE_NAME, kernel_path);
+
+        if has_rd_path {
+            files.insert(RAMDISK_FILE_NAME, ramdisk_path);
+        }
```
phil-opp, line 161: Same as above (use `if let` and avoid the additional bindings in lines 151-154).

src/lib.rs
```diff
@@ -17,19 +17,30 @@ mod mbr;
...
+    }
+
+    /// Add a ramdisk file to the image
+    pub fn with_ramdisk(&self, ramdisk_path: &Path) -> Self {
+        Self {
+            kernel: self.kernel.to_owned(),
+            ramdisk: Some(ramdisk_path.to_owned()),
         }
     }
```
phil-opp, line 45: My preferred style of the builder pattern is to take and return `&mut self`:

```suggestion
    /// Add a ramdisk file to the image
    pub fn set_ramdisk(&mut self, ramdisk_path: &Path) -> &mut Self {
        self.ramdisk = Some(ramdisk_path.to_owned());
        self
    }
```

This way, you can use it either through `BiosBoot::new(..).set_ramdisk(..).create_disk_image(..)` or through:

```rust
let mut boot = UefiBoot::new(..);
boot.set_ramdisk(..);
boot.create_disk_image(..);
```

src/lib.rs
```diff
@@ -78,13 +95,23 @@ impl BiosBoot {
...
+        }
+    }
+
+    /// Add a ramdisk file to the disk image
+    pub fn with_ramdisk(&self, ramdisk_path: &Path) -> Self {
+        Self {
+            kernel: self.kernel.clone(),
+            ramdisk: Some(ramdisk_path.to_owned()),
         }
     }
```
phil-opp, line 116: Same as above: `set_ramdisk(&mut self, ramdisk_path: &Path) -> &mut Self`.

phil-opp (commented):
Thanks for submitting, looks good overall! I left some inline comments below.

phil-opp:
This change only makes sense in combination with https://github.com/rust-osdev/bootloader/pull/302, so maybe we should close this PR and use #302 for the combined changes?

Issue #299, opened 2022-12-20
Support for other architectures 


JakubGawron1:
Hi. Overall, I am impressed with this project, and I am in great admiration. But I have a question, are there any plans to create a bootloader for other processor architectures, e.g. AARCH64, RISC-V? I'm not saying right away that in this repository, maybe a separate bootloader in a separate repository? If such an idea was possible, could I help in some way?
Thank you in advance for your answer and best regards. 😊

phil-opp:
I don't have plans to port the crate to other CPU architectures right now, but I would like to see this as well!

As far as I know, @tsoutsman is currently working on an [UEFI bootloader](https://github.com/tsoutsman/uefi-bootloader) for the [Theseus OS](https://www.theseus-os.com/) project with the goal to support both x86_64 and aarch64 booting. Maybe they can use some help?

(As a side note, I would love to integrate this into the `bootloader` crate at some point if possible.)

tsoutsman:
Yea, the UEFI bootloader supports `x86_64` and should support `aarch64` fairly soon.

I would love to upstream the changes (especially considering a lot of the code is taken from `bootloader`, which I have yet to give credit to), but quite a few of the changes are Theseus-specific.

The big changes that were necessary for the bootloader to work on Theseus were:
- Respecting the load address of the kernel segments
- Adding support for arbitrary files in the UEFI image and providing information about them in the boot info
- Adding kernel ELF sections to the boot info

Only the first point would be a significant hurdle to overcome, which is more a problem with Theseus than the bootloader. Ideally, I'd make Theseus work when loaded at an arbitrary physical address, at which point I could port the changes from `uefi-bootloader` to this repo.

> If such an idea was possible, could I help in some way?

The way I approached it in `uefi-bootloader` is I took all the `x86_64` specific code and placed it in a separate [`arch`](https://github.com/tsoutsman/uefi-bootloader/tree/master/uefi-bootloader/src/arch) directory (akin to `std::sys`). Then, depending on the `target_arch`, [`mod.rs`](https://github.com/tsoutsman/uefi-bootloader/blob/master/uefi-bootloader/src/arch/mod.rs) exports a different module. [`arch/unsupported`](https://github.com/tsoutsman/uefi-bootloader/tree/master/uefi-bootloader/src/arch/unsupported) contains the skeleton API that the other modules should export.

You could start by doing a similar thing for this repo.

phil-opp:
> - Respecting the load address of the kernel segments

[...]

PR #302, opened 2022-12-30
Load ramdisk feature

### LINKED ISSUE #308: Add support for loading ramdisk
Some use cases require additional files to boot.
These can be anything from driver modules, to microkernel early boot services.

The traditional way, and the way things like grub handle this is to load a file and pass it to the kernel.
The bootloader does not interact directly with parsing the file.

I propose we implement this.

jasoncouture:
Adds support for loading a ramdisk to both bios and UEFI, and passing it to the kernel.

Closes #308 
cc #299


bios/stage-2/src/main.rs
```diff
@@ -151,17 +173,21 @@ fn start(disk_number: u16, partition_table_start: *const u8) -> ! {
...
     let disk_buffer_size = disk_buffer.buffer.len();
-    let file = fs
-        .find_file_in_root_dir(file_name, disk_buffer)
-        .expect("file not found");
+    let optional_file = fs.find_file_in_root_dir(file_name, disk_buffer);
+    if optional_file.is_none() {
+        return None;
+    }
+
+    let file = optional_file.unwrap();
```
phil-opp, line 189: We can use a `match` here to avoid the `unwrap`:

```suggestion
    let optional_file = fs.find_file_in_root_dir(file_name, disk_buffer);
    let file = match optional_file {
        Some(file) => file,
        None => return None,
    };
```

Alternatively, we might even be able to use the question mark operator here:
```suggestion
    let file = fs.find_file_in_root_dir(file_name, disk_buffer)?;
```

(not sure if this works, I haven't tested it)

phil-opp:
Thanks a lot for submitting, this is quite a useful feature!

I left some review comments on https://github.com/rust-osdev/bootloader/pull/301, but I think we can apply them directly here instead (and close #301).

There are some build errors on the CI that we should fix. Looks like they're related to the config parsing and some formatting errors (run `cargo fmt` to fix them).

To ensure that the ramdisk loading works as expected, we should also add some tests for it. For example, a test kernel could load some dummy file with some random data and compare its contents after loading.

phil-opp:
Another idea came up in https://github.com/rust-osdev/bootloader/issues/299: Maybe we should support loading a list of files, instead of just a single ramdisk? We could implement this in different ways:

- Place the files in a folder on the FAT partition and let the bootloader load all items of that folder. This would require us to implement FAT directory traversal in our real mode code, though, which might be difficult.
- Concat all files into a single container file in `create_disk_image` and unpack that file in the bootloader before doing the virtual memory mapping. For example, we could make the ramdisk a FAT partition too.

What do you think about this?

jasoncouture:
> Another idea came up in #299: Maybe we should support loading a list of files, instead of just a single ramdisk? We could implement this in different ways:
> 
> * Place the files in a folder on the FAT partition and let the bootloader load all items of that folder. This would require us to implement FAT directory traversal in our real mode code, though, which might be difficult.
> * Concat all files into a single container file in `create_disk_image` and unpack that file in the bootloader before doing the virtual memory mapping. For example, we could make the ramdisk a FAT partition too.
> 
> What do you think about this?
> 
> **Edit:**
> 
> As a third option, we could also keep a single ramdisk file for now and let the kernel (and its build script) handle the packing and unpacking itself.

I see a few issues with loading multiple files:
1. It's unique, and less portable (To say, multiboot2)
2. This, as mentioned is likely to be a challenge for the real mode code (but not impossible)

As for building an archive and parsing/loading with the bootloader:
1. We can build support for archives on top of this change.
2. The kernel may or may not want the archive parsed ahead of time.
3. We'd have to pick a specific format, which could increase coupling of kernels to bootloaders
4. The specific format may also not be appropriate for a particular kernel (IE: newc CPIO, in a kenrel that uses a non-unix like security model)

IMO the arbitrary file loading is useful, but will over complicate the bios portion of the bootloader. And should be skipped.

The archive method would perhaps be a good choice, but can easily be built as a separate, optional feature on top of this change.

Should we just go with the 3rd option? Fix the comments on this PR and use as-is?

jasoncouture:
The tests hang, but the test kernels boot. I'm not sure why.
I didn't really make any changes to the tests, other than allowing them to add a ramdisk.

And it works with https://github.com/jasoncouture/oxidized with and without [this commit](https://github.com/jasoncouture/oxidized/commit/4ccb2fc4ada82479975b958f47db15ef8890e8ac)

phil-opp:
As noted in https://github.com/rust-osdev/bootloader/issues/299#issuecomment-1370796087, I'm fine with the single-file approach for now. The `cpio_reader` crate looks very promising indeed!

We still might want to add native support for multiple files in the future because it still has some advantages:

- It's easier for users to just specify a list of files that should be loaded, without needing to manuall pack/unpack archives.
- It enables more flexible mappings. For example, kernels could configure specific memory offsets for different files.
- Compined with ASLR, it would increase security if we map the individual files at different offsets. This way, attackers could no longer deduce the addresses of other files from a file's address.

bios/stage-2/src/main.rs
```diff
@@ -27,6 +27,7 @@ const BOOTLOADER_SECOND_STAGE_PARTITION_TYPE: u8 = 0x20;
 const STAGE_3_DST: *mut u8 = 0x0010_0000 as *mut u8; // 1MiB (typically 14MiB accessible here)
 const STAGE_4_DST: *mut u8 = 0x0020_0000 as *mut u8; // 2MiB (typically still 13MiB accessible here)
 const KERNEL_DST: *mut u8 = 0x0100_0000 as *mut u8; // 16MiB
+const RAMDISK_DST: *mut u8 = 0x0400_0000 as *mut u8; // 64MiB
```
phil-opp, line 30: Hardcoding an address here could be problematic because it effectively limits the possible kernel size.

bios/stage-2/src/main.rs
```diff
@@ -98,6 +99,17 @@ fn start(disk_number: u16, partition_table_start: *const u8) -> ! {
     writeln!(screen::Writer, "loading kernel...").unwrap();
     let kernel_len = load_file("kernel-x86_64", KERNEL_DST, &mut fs, &mut disk, disk_buffer);
     writeln!(screen::Writer, "kernel loaded at {KERNEL_DST:#p}").unwrap();
+    writeln!(screen::Writer, "Loading ramdisk...").unwrap();
+    let ramdisk_len = match try_load_file("ramdisk", RAMDISK_DST, &mut fs, &mut disk, disk_buffer) {
```
phil-opp, line 103: Instead of using a hardcoded address for the ramdisk, we could use `KERNEL_DST.wrapping_add(kernel_len)`, rounded up to the next mulitple of teh page size.

bios/stage-4/src/main.rs
```diff
@@ -56,13 +56,21 @@ pub extern "C" fn _start(info: &mut BiosInfo) -> ! {
         PhysAddr::new(info.kernel.start)
     };
     let kernel_size = info.kernel.len;
-    let mut frame_allocator = {
+    let mut frame_allocator = if info.ramdisk.start == 0 {
```
phil-opp, line 59: Right now we always set the `start` field in `stage-2/src/main.rs`, so this condition is never true. We should check the `ramdisk_len` instead.

bios/stage-4/src/main.rs
```diff
@@ -56,13 +56,21 @@ pub extern "C" fn _start(info: &mut BiosInfo) -> ! {
...
         let kernel_end = PhysFrame::containing_address(kernel_start + kernel_size - 1u64);
         let next_free = kernel_end + 1;
         LegacyFrameAllocator::new_starting_at(
             next_free,
             memory_map.iter().copied().map(MemoryRegion),
         )
+    } else {
+        let ramdisk_end =
+            PhysFrame::containing_address(PhysAddr::new(info.ramdisk.start + info.ramdisk.len));
```
phil-opp, line 68: I think we should subtract 1 here to get an inclusive bound.

```suggestion
            PhysFrame::containing_address(PhysAddr::new(info.ramdisk.start + info.ramdisk.len - 1));
```

uefi/src/main.rs
```diff
@@ -63,39 +69,77 @@ fn efi_main(image: Handle, st: SystemTable<Boot>) -> Status {
 
 fn main_inner(image: Handle, mut st: SystemTable<Boot>) -> Status {
     // temporarily clone the y table for printing panics
+
     unsafe {
         *SYSTEM_TABLE.get() = Some(st.unsafe_clone());
     }
+    let stdout = unsafe { &mut *SYSTEM_TABLE.get() };
+    let stdout = stdout.as_mut().unwrap();
```
phil-opp, line 77: The only reason for the `SYSTEM_TABLE` static is that we can print on panics. We can use `st` directly here.

```suggestion
```

uefi/src/main.rs
```diff
@@ -63,39 +69,77 @@ fn efi_main(image: Handle, st: SystemTable<Boot>) -> Status {
...
+    let stdout = stdout.stdout();
+    stdout.clear().unwrap();
```
phil-opp, line 79: We don't want to hold on to the `&mut` stdout reference in case a panic happens. So instead of defining a `stdout` binding, we want to use `st.stdout()` directly. This also applies to the `writeln!` calls below.

```suggestion
    st.stdout().clear().unwrap();
```

uefi/src/main.rs
```diff
@@ -63,39 +69,77 @@ fn efi_main(image: Handle, st: SystemTable<Boot>) -> Status {
...
+    writeln!(stdout, "UEFI bootloader started; trying to load kernel").unwrap();
```
phil-opp, line 80:
```suggestion
    writeln!(st.stdout(), "UEFI bootloader started; trying to load kernel").unwrap();
```

uefi/src/main.rs
```diff
@@ -366,11 +448,22 @@ fn create_page_tables(
...
-        .locate_protocol::<GraphicsOutput>()
+        .get_handle_for_protocol::<GraphicsOutput>()
         .ok()?;
-    let gop = unsafe { &mut *gop.get() };
+    let mut gop = unsafe {
+        st.boot_services()
+            .open_protocol::<GraphicsOutput>(
+                OpenProtocolParams {
+                    handle: gop_handle,
+                    agent: st.boot_services().image_handle(),
```
phil-opp, line 460: Is this the correct handle? Or should we use the `Handle` passed to `main_inner`?

uefi/src/main.rs
```diff
@@ -130,173 +182,203 @@ fn main_inner(image: Handle, mut st: SystemTable<Boot>) -> Status {
...
-    let base_code = unsafe { &mut *base_code_raw.interface.get() };
+    let mut base_code_raw = locate_and_open_protocol::<BaseCode>(image, st)?;
+    let base_code = base_code_raw.deref_mut();
 
     // Find the TFTP boot server.
     let mode = base_code.mode();
     assert!(mode.dhcp_ack_received);
     let dhcpv4: &DhcpV4Packet = mode.dhcp_ack.as_ref();
     let server_ip = IpAddress::new_v4(dhcpv4.bootp_si_addr);
+    let mut buf = [0u8; 256];
```
phil-opp, line 356: This appears to be unused.

```suggestion
```

uefi/src/main.rs
```diff
@@ -130,173 +182,203 @@ fn main_inner(image: Handle, mut st: SystemTable<Boot>) -> Status {
...
+    let filename = CStr16::from_str_with_buf(name.trim_end_matches('\0'), &mut buf)
+        .expect("Failed to convert string to utf16");
+
+    let file_handle_result = root.open(filename, FileMode::Read, FileAttribute::empty());
+
+    if file_handle_result.is_err() {
+        return None;
+    }
+
+    let file_handle = file_handle_result.unwrap();
```
phil-opp, line 316: We can use a match here to avoid the `unwrap()`:

```suggestion
    let file_handle = match file_handle_result {
        Err(_) => return None,
        Some(handle) = handle,
    };
```

src/lib.rs
```diff
@@ -17,22 +17,31 @@ mod mbr;
...
         }
     }
 
+    /// Add a ramdisk file to the image
+    pub fn set_ramdisk(&mut self, ramdisk_path: &Path) -> &mut Self {
+        self.ramdisk = Some(ramdisk_path.to_owned());
+        self
+    }
+
     /// Create a bootable UEFI disk image at the given path.
```
asensio-project, line 45: This comment is inconsistent because this function creates BIOS images instead UEFI ones.
I suggest you to change ``UEFI disk image`` to ``BIOS disk image``

jasoncouture: Will do. Thank you!

uefi/src/main.rs
```diff
@@ -63,39 +69,77 @@ fn efi_main(image: Handle, st: SystemTable<Boot>) -> Status {
...
+                let next_target_size = memory_map_size.map_size + (8 * memory_map_size.entry_size);
+                target_size = next_target_size;
+                st.boot_services()
+                    .free_pool(ptr)
+                    .expect("Failed to free temporary memory for memory map!");
+                continue;
+            }
+            break;
+        }
+        storage
```
phil-opp, line 142: Could you clarify why this change was needed? Did you run into any errors because of the previous implementation?

phil-opp, line 142: We can simplify this loop a bit:

```suggestion
        loop {
            let ptr = st
                .boot_services()
                .allocate_pool(MemoryType::LOADER_DATA, target_size)
                .expect("Failed to allocate memory for mmap storage");
            let storage = unsafe { slice::from_raw_parts_mut(ptr, target_size) };
            if st.boot_services().memory_map(storage).is_ok() {
                break storage;
            }
            // allocated memory region was not big enough -> free it again
            st.boot_services()
                .free_pool(ptr)
                .expect("Failed to free temporary memory for memory map!");

            // By measuring the size here, we can find out exactly how much we need.
            // We may hit this code twice, if the map allocation ends up spanning more pages.
            memory_map_size = st.boot_services().memory_map_size();

            let next_target_size = memory_map_size.map_size + (8 * memory_map_size.entry_size);
            target_size = next_target_size;
            
        }
```

jasoncouture: The measurement needs to be done prior to freeing the pool, this removes the need for the + 8 entries.

jasoncouture:
I'll fix the build errors when I rebase this later. I'm at work and wanted to get them in so I don't forget.

common/src/lib.rs
```diff
@@ -267,6 +270,39 @@ where
...
+        let physical_address = PhysAddr::new(ramdisk_address);
+        let ramdisk_physical_start_page: PhysFrame<Size4KiB> =
+            PhysFrame::containing_address(physical_address);
+        let ramdisk_page_count = (system_info.ramdisk_len - 1 / Size4KiB::SIZE) + 1;
+        let ramdisk_physical_end_page = ramdisk_physical_start_page + ramdisk_page_count;
+        let start_page = Page::containing_address(ramdisk_address_start);
+
+        let flags = PageTableFlags::PRESENT | PageTableFlags::WRITABLE;
+        for (i, frame) in
+            PhysFrame::range_inclusive(ramdisk_physical_start_page, ramdisk_physical_end_page)
```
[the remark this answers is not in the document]
jasoncouture: I updated the match to get rid of the russian nesting doll of parentheses instead.

uefi/src/main.rs
jasoncouture: I verified with the docs again.

Unfortunately the handle is opaque, and does not implement the method needed to compare them, so I couldn't assert.

jasoncouture: Document comments:
```
    /// Get the [`Handle`] of the currently-executing image.
    pub fn image_handle(&self) -> Handle {
        // Safety:
        //
        // `IMAGE_HANDLE` is only set by `set_image_handle`, see that
        // documentation for more details.
        //
        // Additionally, `image_handle` takes a `&self` which ensures it
        // can only be called while boot services are active. (After
        // exiting boot services, the image handle should not be
        // considered valid.)
```

And :

```
    /// Update the global image [`Handle`].
    ///
    /// This is called automatically in the `main` entry point as part
    /// of [`uefi_macros::entry`]. It should not be called at any other
    /// point in time, unless the executable does not use
    /// [`uefi_macros::entry`], in which case it should be called once
    /// before calling other `BootServices` functions.
```

phil-opp:
Looks like there are still some unaddressed comments, that GitHub helpfully hides by default :D

image

common/src/lib.rs
```diff
@@ -267,6 +270,40 @@ where
...
+        let ramdisk_address_start = mapping_addr(
+            config.mappings.ramdisk_memory,
+            system_info.ramdisk_len,
+            Size4KiB::SIZE,
+            &mut used_entries,
+        );
+        let physical_address = PhysAddr::new(ramdisk_address);
+        let ramdisk_physical_start_page: PhysFrame<Size4KiB> =
+            PhysFrame::containing_address(physical_address);
+        let ramdisk_page_count = system_info.ramdisk_len / Size4KiB::SIZE;
```
phil-opp, line 284: We still need the first `-1` because otherwise we allocate one page more than needed if the size is exactly a multiple of `Size4KiB::SIZE`.

```suggestion
        let ramdisk_page_count = (system_info.ramdisk_len - 1) / Size4KiB::SIZE;
```

uefi/src/main.rs
phil-opp: Thanks for clarifying! I think I'd still prefer if we just passed the handle from main as an addiitional argument. It's more obvious this way imo so there is less chance for confusion when we modify this part of the code again in the future.

uefi/src/main.rs
phil-opp: Ok, thanks for clarifying! The new code is more resilient than before since it retries the allocation instead of giving up directly, so I'm fine with this change.

jasoncouture:
> Looks like there are still some unaddressed comments, that GitHub helpfully hides by default :D
> 
> image

Ah yes, because it's a great usability feature. 😁 
I'll take care of it today.

uefi/src/main.rs
jasoncouture: This was originally done during test debugging where it appeared boot was hanging (But was a TFTP boot failure).

Without grabbing the SYSTEM_TABLE pointer, and getting stdout from there, I couldn't figure out how to get it compile :sweat_smile: 

Fixed.

uefi/src/main.rs
jasoncouture: I think it'd be less confusing to go the other way. It also requires less modifications in the future when UEFI features are added (As we wouldn't be passing the handle through the call stack).

So, IMO, maybe it'd be better to go the other way, and stop using handle directly, instead using it from boot services. But that will require more discussion, and obviously I'll stick with which direction you want to go.

Change has been made, as requested. Regardless of path, it's definitely not great to be inconsistent. :)

jasoncouture:
I've rebased on main @phil-opp , I think this is ready to be merged.

jasoncouture:
In order to resolve the merge conflict resulting from the merge of #304 , I'm going to need to revert it, and instead merge in #313

phil-opp:
> In order to resolve the merge conflict resulting from the merge of #304 , I'm going to need to revert it, and instead merge in #313

Oh wow, now I'm confused.. This should not be necessary, we just need to fix a few conflicts when merging `main` into 64faf0faa59535937a047c48c1c6e8ad66b89f41. I can take care of these conflicts if you like, but the reverting, the integration of #313, and the force-pushes make reviewing this PR much more difficult.

So my preferred approach would be to force-push this PR back to 64faf0faa59535937a047c48c1c6e8ad66b89f41. Then I can give it another round of review and resolve the conflicts.

jasoncouture:
> 

Understood.

> > In order to resolve the merge conflict resulting from the merge of #304 , I'm going to need to revert it, and instead merge in #313
> 
> Oh wow, now I'm confused.. This should not be necessary, we just need to fix a few conflicts when merging `main` into [64faf0f](https://github.com/rust-osdev/bootloader/commit/64faf0faa59535937a047c48c1c6e8ad66b89f41). I can take care of these conflicts if you like, but the reverting, the integration of #313, and the force-pushes make reviewing this PR much more difficult.
> 
> So my preferred approach would be to force-push this PR back to [64faf0f](https://github.com/rust-osdev/bootloader/commit/64faf0faa59535937a047c48c1c6e8ad66b89f41). Then I can give it another round of review and resolve the conflicts.

jasoncouture:
> > 
> 
> Understood.
> 
> > > In order to resolve the merge conflict resulting from the merge of #304 , I'm going to need to revert it, and instead merge in #313
> > 
> > 
> > Oh wow, now I'm confused.. This should not be necessary, we just need to fix a few conflicts when merging `main` into [64faf0f](https://github.com/rust-osdev/bootloader/commit/64faf0faa59535937a047c48c1c6e8ad66b89f41). I can take care of these conflicts if you like, but the reverting, the integration of #313, and the force-pushes make reviewing this PR much more difficult.
> > So my preferred approach would be to force-push this PR back to [64faf0f](https://github.com/rust-osdev/bootloader/commit/64faf0faa59535937a047c48c1c6e8ad66b89f41). Then I can give it another round of review and resolve the conflicts.

Headed for [0a4b9c7](https://github.com/rust-osdev/bootloader/commit/0a4b9c719506774cbbbe2a4379589b0d0445aa9b) instead, as that was merged on top of the logging changes.

jasoncouture:
Changes from the logger merge are here: https://github.com/rust-osdev/bootloader/compare/16fd0bb04da7975e0ea2438db6885df9064444a5..0a4b9c719506774cbbbe2a4379589b0d0445aa9b

phil-opp (approved):
Looks good to me, thanks a lot for all the updates!

I'll handle the conflicts and merge this PR afterwards.

PR #303, opened 2022-12-30
Log level feature

### LINKED ISSUE #288: Add a QUIET feature

asensio-project:
Hello,

This PR adds a log_level configuration in bootloader_api. Now, the developers of the kernel can change which kind of messages are shown in the screen when booting.

Thanks for this project.

(This is the replacement of #292)

Resolves #288

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -9,7 +9,6 @@ description = "Makes a kernel compatible with the bootloader crate"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-log = "0.4.17"
 
 [dev-dependencies]
 rand = "0.8.4"
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -9,6 +9,7 @@ description = "Makes a kernel compatible with the bootloader crate"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+log = "0.4.17"
 
 [dev-dependencies]
 rand = "0.8.4"
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -149,10 +149,11 @@ impl BootloaderConfig {
 
         let log_level: u8 = match log_level {
             log::LevelFilter::Trace => 0,
-            log::LevelFilter::Info => 1,
-            log::LevelFilter::Warn => 2,
-            log::LevelFilter::Error => 3,
-            log::LevelFilter::Off => 4,
+            log::LevelFilter::Debug => 1,
+            log::LevelFilter::Info => 2,
+            log::LevelFilter::Warn => 3,
+            log::LevelFilter::Error => 4,
+            log::LevelFilter::Off => 5,
         };
 
         concat_115_1(buf, log_level.to_le_bytes())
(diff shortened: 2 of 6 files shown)


api/src/config.rs
```diff
@@ -133,13 +139,24 @@ impl BootloaderConfig {
...
+        );
+
+        let log_level: u8 = match log_level {
+            log::LevelFilter::Trace => 0,
+            log::LevelFilter::Debug => 1,
+            log::LevelFilter::Info => 2,
+            log::LevelFilter::Warn => 3,
+            log::LevelFilter::Error => 4,
+            log::LevelFilter::Off => 5,
+        };
```
phil-opp, line 157: The `LevelFilter` enum is `repr(usize)`, so I don't think that it's a good idea to use a different integer representation. Instead, we should use the canonical one:

```suggestion
        let log_level = u8::try_from(log_level as usize).unwrap();
```

If we define our own `LevelFilter` enum, we could even make it `repr(u8)` to avoid the `u8::try_from`.

api/Cargo.toml
```diff
@@ -9,6 +9,7 @@ description = "Makes a kernel compatible with the bootloader crate"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+log = "0.4.17"
```
phil-opp, line 12: The `log` crate defines some global items, which might lead to problems when the user's kernel also depends on a different version of the `log` crate. Since we only need the `LevelFilter` enum I think it's better to define our own copy of that enum and avoid this dependency.

api/src/config.rs
```diff
@@ -252,6 +269,17 @@ impl BootloaderConfig {
...
+        let (&log_level, s) = split_array_ref(s);
+        let log_level = match u8::from_le_bytes(log_level) {
+            0 => log::LevelFilter::Trace,
+            1 => log::LevelFilter::Debug,
+            2 => log::LevelFilter::Info,
+            3 => log::LevelFilter::Warn,
+            4 => log::LevelFilter::Error,
+            5 => log::LevelFilter::Off,
+            _ => return Err("log_level invalid"),
+        };
```
phil-opp, line 281: When we create our own `LevelFilter` enum, we should include a `from_u8(value: u8) -> Option<LevelFilter>` method and use it here (so that we have the integer representation in a single place).

phil-opp (commented):
Thanks for the PR! Looks good to me overall, but I would prefer to create our own copy of the `LevelFilter` enum instead of adding a `log` dependency to the `bootloader_api` crate (see my comments below).

api/src/config.rs
asensio-project: Thanks for the feedback! 
One question: 
So the kernel must use our implementation of the LevelFilter instead the log's one.
And also we have to make a casting in the uefi and bios crate to convert our LevelFilter into the log's LevelFilter.
Is that what you are saying?

asensio-project: The method from_usize() in log crate is a private function. So we need another function (maybe in the common crate) that converts our LevelFilter to u8 , the u8 to usize and a match statement for converting the usize into the log's LevelFitler. And then passing the result to set_max_level().

phil-opp: Yeah, what I meant is that we should duplicate the `LevelFilter` enum of the log crate, together with its `from_usize` method. We can then change `usize` to `u8` to avoid some casts.

> So we need another function (maybe in the common crate) that converts our LevelFilter to [...] log's LevelFitler.

Yeah, we need to do this conversion. The simplest way to implement it is probably something like the following:


```rust
fn convert_level(level: LevelFilter) -> log::LevelFilter {
    match level {
        LevelFilter::Error => log::LevelFilter::Error,
        LevelFilter::Warn => log::LevelFilter::Warn,
        LevelFilter::Info => log::LevelFilter::Info,
        [...]
    }
}

asensio-project:
Since ``d1da62f``, this PR doesn't depend in the log crate for the API.

phil-opp (approved):
Looks good, thanks for the quick update!

PR #304, opened 2023-01-02
 Add bios and uefi cargo features (closes #287)

### LINKED ISSUE #287: Add cargo features to build only UEFI or BIOS bootloader
Some projects only need one build standard. So we should allow to build only the UEFI or the BIOS bootloader to make it possible to reduce compile times. To achieve this, we could add new `uefi` and `bios` cargo features, which are both enabled by default

Proposed by @AviiNL [on gitter](https://gitter.im/rust-osdev/Lobby?at=63809adcbcdb0060b839bde3).

AlexJMohr:
closes #287 

- Added `bios` and `uefi` features in `Cargo.toml` (both enabled by default)
- Made `mbrman` and `gpt` dependencies optional, enabled by the `bios` and `uefi` features respectively
- Moved `BiosBoot` to new `src/bios/mod.rs`
- Moved `mbr.rs` to `src/bios` since it is only used in bios boot
- Moved `UefiBoot` to new `src/uefi/mod.rs`
- Moved `gpt.rs` and `pxe.rs` to `src/uefi` since they are only used in uefi boot
- publicly export `BiosBoot` and `UefiBoot` in `lib.rs`, gated by the `bios` and `uefi` features respectively
- Update `build.rs` to use the new features to conditionally compile bios or uefi related crates


build.rs
```diff
@@ -273,23 +270,23 @@ fn convert_elf_to_bin(elf_path: PathBuf) -> PathBuf {
...
-#[cfg(docsrs_dummy_build)]
+#[cfg(any(docsrs_dummy_build, not(feature = "bios")))]
 fn build_bios_stage_3(_out_dir: &Path) -> PathBuf {
     PathBuf::new()
 }
-#[cfg(docsrs_dummy_build)]
+#[cfg(any(docsrs_dummy_build, not(feature = "bios")))]
 fn build_bios_stage_4(_out_dir: &Path) -> PathBuf {
     PathBuf::new()
 }
```
phil-opp, line 292: Why do we need to modify the feature gates here?

phil-opp:
Thanks a lot! I only took a quick look, but what I saw looked good. I try to do a full review in the next few days.

build.rs
AlexJMohr: If one of the features is disabled, for example `uefi`, then `build_uefi_bootloader` is undefined. It is used in the `main` function in build.rs.

A different solution would be to do this, so the `build_*_bootloader` functions can be left undefined:
```rs
    #[cfg(feature = "uefi")]
    {
        let uefi_path = build_uefi_bootloader(&out_dir);
        println!(
            "cargo:rustc-env=UEFI_BOOTLOADER_PATH={}",
            uefi_path.display()
        );
    }

    #[cfg(feature = "bios")]
    {
        let bios_boot_sector_path = build_bios_boot_sector(&out_dir);
        println!(
            "cargo:rustc-env=BIOS_BOOT_SECTOR_PATH={}",
            bios_boot_sector_path.display()
        );
        let bios_stage_2_path = build_bios_stage_2(&out_dir);
        println!(
            "cargo:rustc-env=BIOS_STAGE_2_PATH={}",
            bios_stage_2_path.display()
        );

        let bios_stage_3_path = build_bios_stage_3(&out_dir);
        println!(
            "cargo:rustc-env=BIOS_STAGE_3_PATH={}",
            bios_stage_3_path.display()
        );

        let bios_stage_4_path = build_bios_stage_4(&out_dir);
        println!(
            "cargo:rustc-env=BIOS_STAGE_4_PATH={}",
            bios_stage_4_path.display()
        );
    }
```

Which do you prefer?

phil-opp: I think I prefer the second variant, i.e. the additional feature gates in `main`. It's easier to understand what's going on this way in my opinion.

jasoncouture:
I have an API change that will conflict with this (Two actually)

The newest refactors the separate BiosBoot/UefiBoot into a single DiskImageBuilder struct/impl.

The purpose of this upcoming change will be to support loading arbitrary files, and a text configuration file.

@AlexJMohr would you mind if I add these changes to that API change, since I'll be breaking things anyway?

jasoncouture:
Here's the API change that I can add this to: https://github.com/rust-osdev/bootloader/commit/154dbabea43d8b733095ab5bd3db7719d9be457b

AlexJMohr:
> @AlexJMohr would you mind if I add these changes to that API change, since I'll be breaking things anyway?

If it were up to me I'd say merge this PR first, then integrate the changes into your branch. But it's more up to @phil-opp on how to handle that.

phil-opp (approved):
Looks good, thanks a lot!

phil-opp:
I pushed two more commits to allow running our test framework only for BIOS or UEFI and for testing the feature gates on the CI.

phil-opp:
Regarding the conflicts with #302 and #313: We will have some conflicts either way and I don't think that one direction is significantly harder than the other. So I don't see any reason to block this PR, which is otherwise ready.

@jasoncouture I'm happy to help with resolving conflicts in #302!

Cargo.toml
```diff
@@ -37,12 +37,17 @@ bootloader_api = { version = "0.11.0", path = "api" }
 bootloader-x86_64-common = { version = "0.11.0", path = "common" }
 bootloader-x86_64-bios-common = { version = "0.11.0", path = "bios/common" }
 
+[features]
+default = ["bios", "uefi"]
+bios = ["mbrman", "bootloader_test_runner/bios"]
+uefi = ["gpt", "bootloader_test_runner/uefi"]
```
Freax13, line 43:
```suggestion
bios = ["dep:mbrman", "bootloader_test_runner/bios"]
uefi = ["dep:gpt", "bootloader_test_runner/uefi"]
```

We should use the `dep:` syntax to enable the dependencies. Otherwise cargo will create and expose features called `mbrman` and `gpt` for the `bootloader` package.

Freax13 (changes requested):

Issue #305, opened 2023-01-03
Unit test improvements


jasoncouture:
During development of #302 tftp boots broke.

This caused the tests to hang, and never display the tftp failure panic.

I propose the following changes:
1. Add support for bootloader panic hander that uses the qemu exit port.
2. Write a macro that splits the tests into _bios, _uefi, and _tftp

phil-opp:
Sounds reasonable to me!

PR #307, opened 2023-01-04
Feature test macros

jasoncouture:
This PR addresses the second item in #305 by separating the various boot methods and ramdisk tests into separate tests.
It will also generate ramdisk tests, unless explicitly told not to.


jasoncouture:
The changes for this PR start at commit 3f24dd4da68dce1546baf11c40c906ed769f5ae6

jasoncouture:
The windows test failed due to timeout, see #309 for a potential fix

phil-opp (commented):
Thanks a lot!

A few points:

- I don't think that we need to duplicate all the tests with and without ramdisk. A few ramdisk-specific tests should be enough, the other tests can use `None`.
  - This should also allow us to simplify the `define_test` macro.
- This PR does not really depend on the ramdisk PR, so we could already apply it on top of `main`. We can of course also wait until #302 is merged and rebase this PR afterwards if you prefer.
- Could you explain the problem with the LTO test in more detail? I'm not sure what's the problem.

PR #313, opened 2023-01-05
Simplified disk builder

### LINKED ISSUE #287: Add cargo features to build only UEFI or BIOS bootloader

jasoncouture:
Part one of addressing issue #299  requires PR #302 first, and incorporates changes from PR #304
Closes #287 

Starts at commit: 154dbabea43d8b733095ab5bd3db7719d9be457b


jasoncouture:
Merged into #302

jasoncouture:
Going to open a new PR for my other branch for this instead. which is compatible with the current main branch.

PR #314, opened 2023-01-05
Improve Logging

### LINKED ISSUE #306: Better logging
The UEFI and bios loggers use only the framebuffer.

This can be an issue during headless runs (server with no display, unit tests)

I propose we refactor the logger to support multiple methods of logging output. (Serial being the first idea), with support in the kernel configuration to enable/disable sources. Or a text config file (this is likely better, so it can be configured at runtime)

This would also allow us to separate the logging and framebuffer init, as well be able to add/remove loggers at will. Rather than using `writeln` for early logging, we will be able to initialize the logger before anything else, and configure it with the optional `config.txt` or a default config that matches the current behavior.

asensio-project:
Hello,

This PR improves the actual logging of the bootloader.

Tasks:

- [x] Separate the logger from the framebuffer
- [x] Add support for Serial Being logger
- [x] Add configurations for enabling or disabling these methods of logging

Closes #306

Thanks!


jasoncouture:
Makes sense. I'll also block `config.txt` on this as well, once the kernel config is in place.
It may yet again be wise to wait on #302 to avoid conflicts before doing the kernel config. So, perhaps do what is currently in this PR separately, as it'll be a huge help.

api/src/config.rs
```diff
@@ -31,6 +31,12 @@ pub struct BootloaderConfig {
     /// Configuration for changing the level of the filter of the messages that are shown in the
     /// screen when booting. The default is 'Trace'.
     pub log_level: LevelFilter,
+
+    /// Configuration for enable or disable the framebuffer logger.
+    pub frame_buffer_logger_status: LoggerStatus,
+
+    /// Configuration for enable or disable the serial being logger.
+    pub serial_logger_status: LoggerStatus,
```
phil-opp, line 39: Minor wording suggestions:

```suggestion

    /// Whether the bootloader should print log messages to the framebuffer when booting.
    ///
    /// Enabled by default.
    pub frame_buffer_logger_status: LoggerStatus,

    /// Whether the bootloader should print log messages to the serial port when booting.
    ///
    /// Enabled by default.
    pub serial_logger_status: LoggerStatus,
```

common/src/logger.rs
```diff
@@ -1,71 +1,45 @@
...
-    }
-    get(c).unwrap_or_else(|| get(BACKUP_CHAR).expect("Should get raster of backup char."))
-}
+/// A [`FrameBufferWriter`] instance protected by a spinlock.
+pub struct LockedLogger(
+    Spinlock<FrameBufferWriter>,
+    Spinlock<SerialPort>,
+    LoggerStatus,
+    LoggerStatus,
+);
```
phil-opp, line 16: I would prefer to make this a normal struct with named fields. Also, we could directy resolve the `LoggerStatus` and set the fields as an `Option`:

```suggestion
/// A logger instance protected by a spinlock.
pub struct LockedLogger {
    framebuffer: OptionSpinlock<FrameBufferWriter>>,
    serial: Option<Spinlock<SerialPort>>,
}
```

common/src/logger.rs
```diff
@@ -1,71 +1,45 @@
...
+        info: FrameBufferInfo,
+        frame_buffer_logger_status: LoggerStatus,
+        serial_logger_status: LoggerStatus,
+    ) -> Self {
+        LockedLogger(
+            Spinlock::new(FrameBufferWriter::new(framebuffer, info)),
+            Spinlock::new(SerialPort::new()),
+            frame_buffer_logger_status,
+            serial_logger_status,
+        )
```
phil-opp, line 31: As mentioned above:

```suggestion
    pub fn new(
        framebuffer: &'static mut [u8],
        info: FrameBufferInfo,
        frame_buffer_logger_status: LoggerStatus,
        serial_logger_status: LoggerStatus,
    ) -> Self {
        let framebuffer = match frame_buffer_logger_status {
            LoggerStatus::Enable => Some(Spinlock::new(FrameBufferWriter::new(framebuffer, info))),
            LoggerStatus::Disable => None,
        };
        let serial = match serial_logger_status {
            LoggerStatus::Enable => Some(SerialPort::new()),
            LoggerStatus::Disable => None,
        };
        LockedLogger {
            framebuffer,
            serial,
        }
```

common/src/logger.rs
```diff
@@ -1,71 +1,45 @@
...
     /// Force-unlocks the logger to prevent a deadlock.
     ///
     /// ## Safety
     /// This method is not memory safe and should be only used when absolutely necessary.
     pub unsafe fn force_unlock(&self) {
-        unsafe { self.0.force_unlock() };
+        unsafe {
+            self.0.force_unlock();
+            self.1.force_unlock();
+        };
```
phil-opp, line 42: With the above changes, we could use `if let` here:

```suggestion
        if let Some(framebuffer) = &mut self.framebuffer {
            unsafe { framebuffer.force_unlock() };
        }
        if let Some(serial) = &mut self.serial {
            unsafe { serial.force_unlock() };
        }
```

common/src/logger.rs
```diff
@@ -75,119 +49,15 @@ impl log::Log for LockedLogger {
...
-                }
-                let new_ypos =
-                    self.y_pos + font_constants::CHAR_RASTER_HEIGHT.val() + BORDER_PADDING;
-                if new_ypos >= self.height() {
-                    self.clear();
-                }
-                self.write_rendered_char(get_char_raster(c));
-            }
+        if self.2 == LoggerStatus::Enable {
+            let mut logger = self.0.lock();
```
phil-opp, line 53:
```suggestion
        if let Some(framebuffer) = &mut self.framebuffer {
            let mut logger = framebuffer.lock();
```

common/src/logger.rs
```diff
@@ -75,119 +49,15 @@ impl log::Log for LockedLogger {
...
-
-    /// Prints a rendered char into the framebuffer.
-    /// Updates `self.x_pos`.
-    fn write_rendered_char(&mut self, rendered_char: RasterizedChar) {
-        for (y, row) in rendered_char.raster().iter().enumerate() {
-            for (x, byte) in row.iter().enumerate() {
-                self.write_pixel(self.x_pos + x, self.y_pos + y, *byte);
-            }
+        if self.3 == LoggerStatus::Enable {
+            let mut serial = self.1.lock();
```
phil-opp, line 57:
```suggestion
        if let Some(serial) = &mut self.serial {
            let mut serial = serial.lock();
```

phil-opp (commented):
Looks very good! I only left some style comments.

asensio-project:
I can't apply the changes proposed by @jasoncouture in the linked issue because it's very difficult to refactor BIOS stages due incompatibility problems of the used crates of the logger. So, I think that this PR is ready to merge.

phil-opp (approved):

phil-opp:
Thanks!

PR #317, opened 2023-01-06
Guard the lower 1MB of memory

### LINKED ISSUE #311: UEFI boot should avoid using the lower 1MB of physical RAM
When activating additional CPUs they will start in real mode.
Trampoline assembly is used to initialize them, after issuing two IPI (inter processor interrupts) (INIT and START) via the APIC.

The start IPI sends the starting code segment for real mode, which is limited to the lower 1MB of memory.

The maximum physical memory address for real mode is FFFFF, which equates to segment F0, offset FFFF

Currently, the bootloader, in UEFI mode, will allocate pages starting near page 0, leaving no space left to place the IPI trampoline.

We could achieve this by allocating all pages below 1MB using UEFI allocate page function in a loop, and freeing them immediately preceding the call to exit boot services. This would ensure the UEFI loader does not allocate and reserve physical pages below 1MB.

jasoncouture:
This also adds a new fuction to ignore the attempts to guard the lower 1MB if it's really required.

Closes #311

### DIFF
--- a/api/src/info.rs
+++ b/api/src/info.rs
@@ -149,12 +149,12 @@ impl MemoryRegion {
 #[non_exhaustive]
 #[repr(C)]
 pub enum MemoryRegionKind {
-    /// Unused conventional memory, can be used by the kernel.
-    Usable,
     /// Memory mappings created by the bootloader, including the page table and boot info mappings.
     ///
     /// This memory should _not_ be used by the kernel.
     Bootloader,
+    /// Unused conventional memory, can be used by the kernel.
+    Usable,
     /// An unknown memory region reported by the UEFI firmware.
     ///
     /// Contains the UEFI memory type tag.
--- a/api/src/info.rs
+++ b/api/src/info.rs
@@ -149,12 +149,12 @@ impl MemoryRegion {
 #[non_exhaustive]
 #[repr(C)]
 pub enum MemoryRegionKind {
+    /// Unused conventional memory, can be used by the kernel.
+    Usable,
     /// Memory mappings created by the bootloader, including the page table and boot info mappings.
     ///
     /// This memory should _not_ be used by the kernel.
     Bootloader,
-    /// Unused conventional memory, can be used by the kernel.
-    Usable,
     /// An unknown memory region reported by the UEFI firmware.
     ///
     /// Contains the UEFI memory type tag.
--- a/common/src/legacy_memory_region.rs
+++ b/common/src/legacy_memory_region.rs
@@ -28,7 +28,10 @@ pub struct LegacyFrameAllocator<I, D> {
     memory_map: I,
     current_descriptor: Option<D>,
     next_frame: PhysFrame,
+    start_frame: PhysFrame,
 }
+
+const LOWER_MEMORY_END_PAGE: u64 = 0x10_0000;
 
 impl<I, D> LegacyFrameAllocator<I, D>
 where
@@ -42,18 +45,32 @@ where
     /// library assumes that references can never point to virtual address `0`.  
     pub fn new(memory_map: I) -> Self {
         // skip frame 0 because the rust core library does not see 0 as a valid address
-        let start_frame = PhysFrame::containing_address(PhysAddr::new(0x1000));
-        Self::new_starting_at(start_frame, memory_map)
+        // also skip the first 1MB of frames, there are use cases that require lower conventional memory access. (Such as SMP SIPI)
+        let start_frame = PhysFrame::containing_address(PhysAddr::new(LOWER_MEMORY_END_PAGE));
+        unsafe { Self::unsafe_new_starting_at(start_frame, memory_map) }
     }
 
     /// Creates a new frame allocator based on the given legacy memory regions. Skips any frames
-    /// before the given `frame`.
+    /// before the given `frame`, or 0x100000, whichever is higher.
     pub fn new_starting_at(frame: PhysFrame, memory_map: I) -> Self {
+        if frame.start_address().as_u64() < LOWER_MEMORY_END_PAGE {
+            // Skip the first 1MB of frames, regardless of what was requested.
+            // there are use cases that require lower conventional memory access. (Such as SMP SIPI)
+            return Self::new(memory_map);
+        }
+
+        return unsafe { Self::unsafe_new_starting_at(frame, memory_map) };
+    }
+
+    /// Creates a new frame allocator based on the given legacy memory regions. Skips any frames
+    /// before the given `frame`, without attempting to protect the lower 1MB of memory.
+    pub unsafe fn unsafe_new_starting_at(frame: PhysFrame, memory_map: I) -> Self {
         Self {
             original: memory_map.clone(),
             memory_map,
             current_descriptor: None,
             next_frame: frame,
+            start_frame: frame,
         }
     }
 
@@ -121,9 +138,11 @@ where
             let next_free = self.next_frame.start_address();
             let kind = match descriptor.kind() {
                 MemoryRegionKind::Usable => {
-                    if end <= next_free {
+                    if end <= next_free && end > self.start_frame.start_address() {
                         MemoryRegionKind::Bootloader
-                    } else if descriptor.start() >= next_free {
+                    } else if descriptor.start() >= next_free
+                        || end <= self.start_frame.start_address()
+                    {
                         MemoryRegionKind::Usable
                     } else {
                         // part of the region is used -> add it separately
--- a/tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
+++ b/tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
@@ -1,0 +1,44 @@
+#![no_std] // don't link the Rust standard library
+#![no_main] // disable all Rust-level entry points
+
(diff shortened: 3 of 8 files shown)


phil-opp (commented):
Thanks a lot for this PR!

I'm not sure if I understand the approach though. Why are the changes to `LegacyMemoryRegion::kind` necessary and how do they help in guarding the first MB?

I think the main issue is that the UEFI allocation used for loading the kernel might already use the first MB if we are unlucky. So to be safe, we would need to allocate that first megabyte before creating the allocation. We can use the "allocate at address" API mentioned by @tsoutsman in https://github.com/rust-osdev/bootloader/issues/311#issuecomment-1372193672 for that.

uefi/src/memory_descriptor.rs
```diff
@@ -20,6 +20,12 @@ impl LegacyMemoryRegion for UefiMemoryDescriptor {
     fn kind(&self) -> MemoryRegionKind {
         match self.0.ty {
             MemoryType::CONVENTIONAL => MemoryRegionKind::Usable,
+            MemoryType::LOADER_CODE => MemoryRegionKind::Usable,
+            // We load the kernel under this type, probably not wise to mark it free after boot, even if it is.
+            // Or perhaps we should use a different memory type (Such as the OS Loader range from the EFI spec (0x80000000-0xFFFFFFFF))?
```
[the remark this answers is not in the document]
jasoncouture: Got it, I'll remove the comment.

api/src/info.rs
```diff
@@ -141,14 +141,14 @@ impl MemoryRegion {
...
-#[repr(C)]
+#[repr(C, u64)]
 pub enum MemoryRegionKind {
     /// Unused conventional memory, can be used by the kernel.
     Usable,
     /// Memory mappings created by the bootloader, including the page table and boot info mappings.
     ///
     /// This memory should _not_ be used by the kernel.
-    Bootloader,
+    Bootloader = 0x8FFFFFFF,
```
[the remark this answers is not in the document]
jasoncouture: It appeared to be conflicting with UEFI boots, since Bootloader and LOADER_CODE both = 1
But that wasn't the case, and may be my misunderstanding of rusts enums. I'll undo this.

uefi/src/memory_descriptor.rs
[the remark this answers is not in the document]
jasoncouture: Actually, I'll leave the comment, and add a note to it with links on why we can't do this.

jasoncouture:
@phil-opp this is ready for review. I also added tests.

common/src/legacy_memory_region.rs
```diff
@@ -28,8 +28,11 @@ pub struct LegacyFrameAllocator<I, D> {
     memory_map: I,
     current_descriptor: Option<D>,
     next_frame: PhysFrame,
+    start_frame: PhysFrame,
```
phil-opp, line 31: We already use the name `start_frame` in multiple places, so we should name this field differently to avoid confusion. I think `min_frame` would be a fitting name:

```suggestion
    min_frame: PhysFrame,
```

common/src/legacy_memory_region.rs
```diff
@@ -42,18 +45,32 @@ where
...
+            // there are use cases that require lower conventional memory access. (Such as SMP SIPI)
+            return Self::new(memory_map);
+        }
+
+        return unsafe { Self::unsafe_new_starting_at(frame, memory_map) };
+    }
+
+    /// Creates a new frame allocator based on the given legacy memory regions. Skips any frames
+    /// before the given `frame`, without attempting to protect the lower 1MB of memory.
+    pub unsafe fn unsafe_new_starting_at(frame: PhysFrame, memory_map: I) -> Self {
```
phil-opp, line 67: This seems a bit complicated. How about we just use a `max` call in `new_starting_at` and keep calling it from the `new` function?

```suggestion
        // Skip the first 1MB of frames, regardless of what was requested.
        // there are use cases that require lower conventional memory access. (Such as SMP SIPI)
        let lower_mem_end = PhysFrame::containing_address(PhysAddr::new(LOWER_MEMORY_END_PAGE));
        let frame = core::cmp::max(frame, lower_mem_end);
```

common/src/legacy_memory_region.rs
```diff
@@ -42,18 +45,32 @@ where
     /// library assumes that references can never point to virtual address `0`.  
     pub fn new(memory_map: I) -> Self {
         // skip frame 0 because the rust core library does not see 0 as a valid address
-        let start_frame = PhysFrame::containing_address(PhysAddr::new(0x1000));
-        Self::new_starting_at(start_frame, memory_map)
+        // also skip the first 1MB of frames, there are use cases that require lower conventional memory access. (Such as SMP SIPI)
+        let start_frame = PhysFrame::containing_address(PhysAddr::new(LOWER_MEMORY_END_PAGE));
+        unsafe { Self::unsafe_new_starting_at(start_frame, memory_map) }
```
phil-opp, line 50: With the suggested changes below, we could keep the previous implementation:

```suggestion
        let start_frame = PhysFrame::containing_address(PhysAddr::new(0x1000));

        Self::new_starting_at(start_frame, memory_map)
```

common/src/legacy_memory_region.rs
```diff
@@ -121,9 +138,11 @@ where
             let next_free = self.next_frame.start_address();
             let kind = match descriptor.kind() {
                 MemoryRegionKind::Usable => {
-                    if end <= next_free {
+                    if end <= next_free && end > self.start_frame.start_address() {
                         MemoryRegionKind::Bootloader
-                    } else if descriptor.start() >= next_free {
+                    } else if descriptor.start() >= next_free
+                        || end <= self.start_frame.start_address()
+                    {
```
phil-opp, line 145: I would prefer to split this condition so that we can add a comment:

```suggestion
                    } else if descriptor.start() >= next_free {
                        MemoryRegionKind::Usable
                    } else if end <= self.start_frame.start_address() {
                        // treat regions that end before the minimum frame as usable too
```

common/src/legacy_memory_region.rs
```diff
@@ -121,9 +138,11 @@ where
...
                         MemoryRegionKind::Usable
                     } else {
                         // part of the region is used -> add it separately
```
phil-opp, line 148: I think we need to update this `else` branch too, in case the descriptor goes beyond `0x100000`.

api/src/info.rs
```diff
@@ -149,12 +149,12 @@ impl MemoryRegion {
...
 #[repr(C)]
 pub enum MemoryRegionKind {
-    /// Unused conventional memory, can be used by the kernel.
-    Usable,
     /// Memory mappings created by the bootloader, including the page table and boot info mappings.
     ///
     /// This memory should _not_ be used by the kernel.
     Bootloader,
+    /// Unused conventional memory, can be used by the kernel.
+    Usable,
```
phil-opp, line 157: Reordering the variants changes their integer value, which is technically a breaking change. So we should undo this change if possible.

tests/test_kernels/lower_memory_free/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+#![no_std]
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum QemuExitCode {
+    Success = 0x10,
+    Failed = 0x11,
+}
+
+pub static RAMDISK_CONTENTS: &[u8] = include_bytes!("../../../ramdisk.txt");
```
phil-opp, line 10: This is never used:

```suggestion
```

tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
```diff
@@ -0,0 +1,44 @@
...
+            "Region: {:016x}-{:016x} - {:?}",
+            region.start,
+            region.end,
+            region.kind
+        )
+        .unwrap();
+        if region.end <= LOWER_MEMORY_END_PAGE && region.kind == MemoryRegionKind::Usable {
+            let pages = (region.end - region.start) / 4096;
+            count += pages;
+        }
```
phil-opp, line 28: We should also account for regions that don't end at `LOWER_MEMORY_END_PAGE`. For example:

```suggestion
        if region.kind == MemoryRegionKind::Usable && region.start < LOWER_MEMORY_END_PAGE {
            let end = core::cmp::min(region.end, LOWER_MEMORY_END_PAGE);
            let pages = (end - region.start) / 4096;
            count += pages;
        }
```

phil-opp:
Thanks for the updates! I left some suggestions, otherwise these changes look good to me. However, I still think that we need to ensure that we don't use the first megabyte for the kernel allocation in the UEFI loader, as I noted above:

> I think the main issue is that the UEFI allocation used for loading the kernel might already use the first MB if we are unlucky. So to be safe, we would need to allocate that first megabyte before creating the allocation. We can use the "allocate at address" API mentioned in https://github.com/rust-osdev/bootloader/issues/311#issuecomment-1372193672 for that.

phil-opp:
@jasoncouture Friendly ping :). What's the status of this?

jasoncouture:
> @jasoncouture Friendly ping :). What's the status of this?

I'll take another look weekend after next. Got some personal things to attend for the next week or so.

Freax13:
Superseded by #446

PR #320, opened 2023-01-09
Simplified disk builder

### LINKED ISSUE #321: Disk builder support for arbitrary files
The disk builders should support the addition of arbitrary files

This is useful for a number of reasons:

1. Smaller OSes may want a single disk, and may be ok with FAT
2. Microkernels may want to load additional services at boot
3. Bootloader file based configuration
4. Ramdisk support
5. Bootloader support for loading arbirary files

Adding all of these files to the same implementation will be cumbersome, due to duplicate of code, so I propose the following:

1. Get rid of UefiBoot and BiosBoot, and replace them with a common builder that has functions for UEFI, BIOS, and TFTP
2. Rather than treating the kernel as a special case, use a BTreeMap on the new implementation to track source/destination

The proposed changes can be found in #320

jasoncouture:
Simplify the disk builder:

1. Move UEFI and BIOS into a common implementation
2. Allow adding of arbitrary files

Closes #321 

Note: No new tests, as existing tests fully cover these changes.


jasoncouture:
@phil-opp this should also be ready to go.

phil-opp:
I only skimmed this PR, but I like the general approach. Some notes:

- Removing the `BiosBoot` and `UefiBoot` structs is a breaking change, so we cannot release this PR in a `v0.11` patch version. So we have two options:
  - Merge this PR once we plan to release `v0.12`.
  - Keep `BiosBoot` and `UefiBoot` and just implement their methods on top of top of the new `DiskImageBuilder`.
- We don't expect that many files on the FAT partition, so I think cloning the `DiskImageFile` fields should be fine. This would allow us to remove the lifetime from the struct and make the whole API simpler.
- Nit: Instead of `&PathBuf`, it's recommended to take `&Path` as argument.
- The changes in `build.rs` seem unrelated to this PR. Removing these changes should also fix the conflict.

jasoncouture:
> I only skimmed this PR, but I like the general approach. Some notes:
> 
> * Removing the `BiosBoot` and `UefiBoot` structs is a breaking change, so we cannot release this PR in a `v0.11` patch version. So we have two options:
>   
>   * Merge this PR once we plan to release `v0.12`.
>   * Keep `BiosBoot` and `UefiBoot` and just implement their methods on top of top of the new `DiskImageBuilder`.

Why not both? Let's keep *Boot, as wrappers on top of `DiskImageBuilder`, and in 0.12 nuke them?

> * We don't expect that many files on the FAT partition, so I think cloning the `DiskImageFile` fields should be fine. This would allow us to remove the lifetime from the struct and make the whole API simpler.

I'll look into it :)
Even if there was a ton of files, do we really need the builder to be high performance?

> * Nit: Instead of `&PathBuf`, it's recommended to take `&Path` as argument.

Will do.

> * The changes in `build.rs` seem unrelated to this PR. Removing these changes should also fix the conflict.

They are a remnant of previous rebases, I'll ditch the changes.

phil-opp:
> Why not both? Let's keep *Boot, as wrappers on top of DiskImageBuilder, and in 0.12 nuke them?

Sure, that would be fine.

> Even if there was a ton of files, do we really need the builder to be high performance?

Faster builds are always nice, but it's not the first priority. What I meant is that it should be fine to call `clone()` in this implementation in order to simplify the API.

jasoncouture:
> Another thought: Maybe we should add an additional builder method to add files in form of a `Vec<u8>` (instead of `&Path`). This would be useful for writing files that are generated on-the-fly without needing to write them to some temporary file first. One example could be some kernel configuration that is serialized using serde.

Agreed. I haven't had a ton of spare time lately, I should be able to get back on this this weekend.

jasoncouture:
>

uefi/src/main.rs
```diff
@@ -1,6 +1,7 @@
 #![no_std]
 #![no_main]
 #![deny(unsafe_op_in_unsafe_fn)]
+#![feature(abi_efiapi)]
```
phil-opp, line 4: This should be no longer required on the latest nightlies.

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
+#[derive(Clone)]
+struct DiskImageFile {
+    source: PathBuf,
+    destination: String,
+}
+
+/// DiskImageBuilder helps create disk images for a specified set of files.
+/// It can currently create MBR (BIOS), GPT (UEFI), and TFTP (UEFI) images.
+pub struct DiskImageBuilder {
+    files: Vec<DiskImageFile>,
```
phil-opp, line 51: Why not use a `BTreeMap` or a `HashMap` here directly?

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
 
 const KERNEL_FILE_NAME: &str = "kernel-x86_64";
 const RAMDISK_FILE_NAME: &str = "ramdisk";
 const CONFIG_FILE_NAME: &str = "boot.json";
+
+#[derive(Clone)]
+struct DiskImageFile {
+    source: PathBuf,
+    destination: String,
+}
```
phil-opp, line 46: We could make the `source` field an enum so that we can represent both files and `Vec<u8>`:

```suggestion
#[derive(Clone)]
struct DiskImageFile {
    source: Source,
    destination: String,
}

enum Source {
    File(PathBuf),
    Bytes(Vec<u8>),
}
```

We can then update `create_fat_filesystem` to use this enum too. This way, we would not need temporary files at all.

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
+            .sync_all()
+            .expect("Failed to flush temp file to disk");
+
+        self.add_or_replace_file(&file_name, target)
+    }
+
+    /// Add or replace arbitrary files.
+    /// NOTE: You can overwrite internal files if you choose, such as EFI/BOOT/BOOTX64.EFI
+    /// This can be useful in situations where you want to generate an image, but not use the provided bootloader.
+    pub fn add_or_replace_file(&mut self, path: &Path, target: &str) -> &mut Self {
```
phil-opp, line 105: How about we name this method just `set_file`? And I would like to swap the argument order here as well (target first, path second).

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
+    /// Configures the runtime behavior of the bootloader.
+    pub fn set_boot_config(&mut self, boot_config: &BootConfig) -> &mut Self {
+        let json =
+            serde_json::to_string_pretty(boot_config).expect("failed to serialize BootConfig");
+        let bytes = json.as_bytes();
+        self.add_or_replace_file_byte_content(bytes, CONFIG_FILE_NAME)
+    }
+
+    /// Add or replace a file from a byte array
+    pub fn add_or_replace_file_byte_content(&mut self, data: &[u8], target: &str) -> &mut Self {
```
phil-opp, line 86: The name seems a bit long. How about `set_file_with_content`? Also, I would prefer to swap the argument order (target path first, data second).

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
+    }
+    fn create_fat_filesystem_image(
+        &self,
+        internal_files: BTreeMap<&str, &Path>,
+    ) -> anyhow::Result<NamedTempFile> {
+        let mut local_map = BTreeMap::new();
+
+        for k in internal_files {
+            local_map.insert(k.0, k.1);
+        }
```
phil-opp, line 123: We should set the internal files after `self.files`, otherwise the user supplied files might overwrite some core files.

Ideally we should also report some error in case there conflicts. Alternatively, we could place all custom files in a subfolder.

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
+
+        let out_file = NamedTempFile::new().context("failed to create temp file")?;
+        fat::create_fat_filesystem(local_map, out_file.path())
+            .context("failed to create BIOS FAT filesystem")?;
+
+        Ok(out_file)
+    }
+    #[cfg(feature = "bios")]
+    /// Create an MBR disk image for booting on BIOS systems.
+    pub fn create_bios_image(&self, image_filename: &Path) -> anyhow::Result<()> {
```
phil-opp, line 137: I don't like the name `image_filename` because it's a full path, not just a file name. Maybe somethingl like `out_path`/`target_path`/`image_path`?

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
+    ) -> anyhow::Result<NamedTempFile> {
+        let mut local_map = BTreeMap::new();
+
+        for k in internal_files {
+            local_map.insert(k.0, k.1);
+        }
+
+        for f in self.files.as_slice() {
+            local_map.insert(&f.destination, f.source.as_path());
+        }
```
phil-opp, line 127: These files are only added to the disk image, but not loaded on boot right? What are your plans about this? Should we try to load these from the bootloader or does the kernel need to create its own disk driver in order to access them.

In case we want to load them in the bootloader, we might need to put additional restrictions on the target file paths. Right now the BIOS bootloader only supports loading files with short file names from the root folder.

src/lib.rs
```diff
@@ -6,18 +6,203 @@ An experimental x86_64 bootloader that works on both BIOS and UEFI systems.
...
+        temp_file
+            .sync_all()
+            .expect("Failed to flush temp file to disk");
+
+        self.add_or_replace_file(&file_name, target)
+    }
+
+    /// Add or replace arbitrary files.
+    /// NOTE: You can overwrite internal files if you choose, such as EFI/BOOT/BOOTX64.EFI
+    /// This can be useful in situations where you want to generate an image, but not use the provided bootloader.
```
phil-opp, line 104: I don't think that we should support this use case. This seems like a footgun, especially because we don't document the names of internal files.

In my opinion it makes more sense to extract the general disk image creation into a separate, indepentent crate. This way, you could use it without building the bootloder as a dependency.

phil-opp (commented):
Thanks for the updates!

src/lib.rs
jasoncouture: hmm, for now I'll remove it, and make the adjustment related to this you suggested. Perhaps this should be in it's own crate.

I'll work on that on my own, happy to hand it over to `rust-osdev` when I'm done. But for now, I think this is big enough as is.

src/lib.rs
jasoncouture: Yes, the files are not loaded at boot.
That would be a future PR to add support, this is just to add the files (Which the kernel can load if it desires).

Still sketching out how the files will get loaded.
Doesn't the FAT library support 8.3 names with LFN? I could look into LFN support at a future date.
Could calling it out in the docs be enough?

src/lib.rs
jasoncouture: Sure!

src/lib.rs
jasoncouture: This was an intentional feature. I will undo it as discussed in the reply below.

As for the error, I'm not sure here personally. I'll go whatever direction you like.
My thought here was: 
During build, we have some default files, we loop and add them.
Later in the build, we have "override" files that we replace these with.

So, with that thought process:
1. Allow overwriting of own files
2. Error when overwriting bootloader files?

src/lib.rs
jasoncouture: Sounds good to me!

src/lib.rs
jasoncouture: Sounds good to me!

src/lib.rs
jasoncouture: The borrow checker mostly.

phil-opp (approved):
Thanks a lot for the updates! Looks good now!

I pushed some more commits with a few minor tweaks.

PR #322, opened 2023-01-09
Cancel in progress PR builds when a new commit is pushed for that PR

jasoncouture:
I've been chewing through actions time on rust-osdev, I'd prefer to be a good citizen so this will cancel in progress builds on PRs when a new commit is pushed.

It does not impact main, or cron builds, due to the fallback to github.ref


phil-opp (approved):
Good idea, thanks!

PR #324, opened 2023-01-09
Implement faster bios builds

jasoncouture:
By executing `llvm-objcopy`, and `cargo-install` in parallel, we can in theory cut bios build time to 25% of it's original build time.

Local testing has taken it from ~1m to ~22s on my Ryzen 7950x

This branch:
`    Finished dev [unoptimized + debuginfo] target(s) in 22.53s`

Main at time of writing:
`    Finished dev [unoptimized + debuginfo] target(s) in 58.25s`


phil-opp (approved):
Good idea, thanks a lot! I can confirm that this change speeds up the build. On my 5900X machine `cargo build` now takes ~32 seconds instead of ~58 seconds.

PR #326, opened 2023-01-14
Add support for a configuration file

asensio-project:
Hello,

This PR adds support for de-serializing a JSON configuration file (``config.json``). I chose JSON because is the only implementation that works in no-std.

- [x] Add a configuration file
- [x] Make it fault tolerant
- [x] Write the tests
- [x] Write/Improve docs

Thanks!


common/Cargo.toml
```diff
@@ -20,6 +20,8 @@ raw-cpuid = "10.2.0"
 rand = { version = "0.8.4", default-features = false }
 rand_hc = "0.3.1"
 uart_16550 = "0.2.18"
+serde-json-core = "0.5.0"
```
bjorn3, line 23: Something like postcard is likely lighter while still supporting serde and no_std.

common/src/config.rs
```diff
@@ -0,0 +1,119 @@
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+pub struct BootloaderConfigFile {
```
bjorn3, line 4: Where is this type serialized? I can't find it.

bjorn3: Or is that part not yet written?

asensio-project: Thanks, my idea was for serializing from the kernel, passing that file to disk image creator and then de-serializing at runtime.

bjorn3: Makes sense

common/Cargo.toml
asensio-project: I will change JSON to postcard, in a future commit. Thanks!

jasoncouture:
Should we keep the kernel configuration as is, and use it as defaults if the config doesn't specify? In lieu of removing those options?

jasoncouture:
Also, shouldnt we load the config first? This would allow the kernel/ramdisk filenames to be set via config if desired (which i personally do desire.)

IE:
* Set config to defaults. 
* If (config_present) load
* for settings that are not set, do not set them leave defaults in place. 
* Load kernel, read kernel config. 
* For each setting not present in config file, apply kernel defaults.
* load ramdisk
* Continue as normal from this point.

We dont need to add the kernel/ramdisk settings now. My goal in these statements is to avoid a breaking api change when we want to add them, and preserve existing behavior.

We can probably even get away with loading the config after, so long as we keep kernel defaults and make the file optional (if it isn't already). We would then be able to refactor the config file as described above, while sticking to a minor semver bump (where as current seems major to be, since it's a breaking change)

common/src/config.rs
```diff
@@ -0,0 +1,125 @@
+use serde::Deserialize;
+use serde_json_core::de;
+
+#[derive(Deserialize)]
+pub struct BootloaderConfigFile {
```
phil-opp, line 5: If we expose this in the API (see below), we should probably rename this to something else (e.g. `BootConfig`).

src/lib.rs
```diff
@@ -18,3 +18,4 @@ pub use uefi::UefiBoot;
 
 const KERNEL_FILE_NAME: &str = "kernel-x86_64";
 const RAMDISK_FILE_NAME: &str = "ramdisk";
+const CONFIG_FILE_NAME: &str = "config.json";
```
phil-opp, line 21: How about naming it `boot.json`? This way, kernels could use their own `config.json` file.

src/bios/mod.rs
```diff
@@ -32,6 +34,12 @@ impl BiosBoot {
         self
     }
 
+    /// Add a JSON configuration file to the disk image
+    pub fn set_config_file(&mut self, config_file_path: &Path) -> &mut Self {
```
phil-opp, line 38: I would prefer to take the `BootloaderConfigFile` struct as argument here. This way, we don't need to write the JSON files by hand and we can statically ensure that the deserialization works.

uefi/src/main.rs
```diff
@@ -97,6 +98,20 @@ fn main_inner(image: Handle, mut st: SystemTable<Boot>) -> Status {
...
+        Err((data, err)) => {
+            writeln!(
+                st.stdout(),
+                "Failed to deserialize the config file {:?}",
+                err
+            )
+            .unwrap();
+            data
+        }
+    };
```
phil-opp, line 113: See above: I think this would be simpler without the `BootloaderConfigFile::deserialize` method.

uefi/src/main.rs
```diff
@@ -107,7 +122,7 @@ fn main_inner(image: Handle, mut st: SystemTable<Boot>) -> Status {
     )
     .unwrap();
 
-    let framebuffer = init_logger(image, &st, kernel.config);
+    let framebuffer = init_logger(image, &st, config);
```
phil-opp, line 125: For compatibility, we need to fall back to the framebuffer configuration of the kernel if no min width/height is set in the config file.

```suggestion
    if config.frame_buffer.minimum_framebuffer_height.is_none() {
        config.frame_buffer.minimum_framebuffer_height = kernel.config.frame_buffer.minimum_framebuffer_height;
    }
    if config.frame_buffer.minimum_framebuffer_width.is_none() {
        config.frame_buffer.minimum_framebuffer_width = kernel.config.frame_buffer.minimum_framebuffer_width;
    }
    let framebuffer = init_logger(image, &st, config);
```

common/src/config.rs
```diff
@@ -0,0 +1,125 @@
...
+    ///
+    /// Enabled by default.
+    #[serde(default)]
+    pub frame_buffer_logger_status: LoggerStatus,
+
+    /// Whether the bootloader should print log messages to the serial port when booting.
+    ///
+    /// Enabled by default.
+    #[serde(default)]
+    pub serial_logger_status: LoggerStatus,
```
phil-opp, line 26: Could we make these fields `bool`s? I don't like the "Enable"/"Disable" in the config file. I would prefer simple `serial_logging`/`frame_buffer_logging` flags.

phil-opp (commented):
Thanks a lot!

phil-opp:
@jasoncouture 

> Should we keep the kernel configuration as is, and use it as defaults if the config doesn't specify? In lieu of removing those options?

We definitely need to keep the frame buffer config in the kernel options for semver compatibility. However, we could mark it as deprecated. Removing the log-related config options is fine because they're not released yet.

> Also, shouldnt we load the config first? This would allow the kernel/ramdisk filenames to be set via config if desired (which i personally do desire.)

We can do that, but I don't think that we need to do it in this PR. Loading it earlier shouldn't be a breaking change.

> IE:
> 
>   - Set config to defaults.
>   - If (config_present) load
>   - for settings that are not set, do not set them leave defaults in place.
>   - Load kernel, read kernel config.
>   - For each setting not present in config file, apply kernel defaults.
>   - load ramdisk
>   - Continue as normal from this point.

I'm not sure if we should make all kernel config options overwritable. The kernel might rely on the config values, e.g. the specific address of a mapping or a minimum stack size. It's probably not a good idea to allow changing these values after compilation.

Cargo.toml
```diff
@@ -49,6 +49,8 @@ fatfs = "0.3.4"
 tempfile = "3.3.0"
 mbrman = { version = "0.5.1", optional = true }
 gpt = { version = "3.0.0", optional = true }
+bootloader-x86_64-common = { version = "0.11.0", path = "common" }
```
phil-opp, line 52: This is quite a big dependency. I think it would be better to make the `common/src/config.rs` a separate crate. We could name it e.g. `bootloader-boot-config` and place it under `common/config`. This way, both the main `bootloader` crate and `bootloader-x86_64-common` could depend on it.

I'm happy to help with this step, just let me know!

phil-opp (changes requested):
Thanks a lot for the updates! Looks very good now.

common/config/src/lib.rs
```diff
@@ -1,6 +1,8 @@
+#![no_std]
+
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Default)]
```
phil-opp, line 5: Deriving `Default` is not what we want here because it will initialize the logging fields with `false` (it doesn't consider the serde-specific attributes). What I meant in https://github.com/rust-osdev/bootloader/pull/326#discussion_r1087736663 was the following:

- keep the manual `Default` implementation
- set the [`serde(default)` attribute](https://serde.rs/container-attrs.html#default) for the `BootConfig` struct
  - This attribute fills any missing fields from the `Default` implementation of the struct.
- remove the `serde(default)` attributes from the struct fields, as they're no longer needed
  - This also removes the need for the `default_logger_status` function.

phil-opp (commented):
Thanks for the quick updates!

common/config/src/lib.rs
asensio-project: Thanks! I didn't understand it. Now, the code is cleaner.

phil-opp (approved):

phil-opp:
Thanks a lot!

PR #328, opened 2023-01-16
Correctness fixes for stage2

### LINKED ISSUE #327: bootloader v0.11 broken under bochs and qemu/kvm
Two issues here: 
* `enter_unreal_mode` should be the very first thing in stage2 main, because the screen write call touches some memory above 64k (`movzx ebx, byte ptr ds:[ecx]` with `ecx=0x13ccc`)
error output in bochs: `00017458470e[CPU0  ] read_virtual_checks(): read beyond limit`

* Rust generates some code that uses `ss` for accesses (`mov eax, dword ptr ss:[ebp+edi*8+4]`, where ss is the default for ebp), and enter_unreal_mode only sets up `ds`. Unsure how this should be fixed properly - set ss to 32 bit as well?
error output in bochs: `00695082304e[CPU0  ] stackPrefetch(4): access [0x00013ff8] > SS.limit [0x0000ffff]`

As for why this works in qemu in emulation mode: qemu is really inaccurate :(

Stary2001:
Correctness fixes to allow the bootloader to work on bochs and kvm.
closes #327

### DIFF
--- a/bios/stage-2/src/main.rs
+++ b/bios/stage-2/src/main.rs
@@ -39,9 +39,10 @@ pub extern "C" fn _start(disk_number: u16, partition_table_start: *const u8) -> 
 }
 
 fn start(disk_number: u16, partition_table_start: *const u8) -> ! {
-    screen::Writer.write_str(" -> SECOND STAGE\n").unwrap();
-
+    // Enter unreal mode before doing anything else.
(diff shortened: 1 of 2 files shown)


phil-opp:
Thanks a lot!

PR #335, opened 2023-01-27
Create kernel stack with correct size and set up a guard page

### LINKED ISSUE #294: Last page of stack memory is not used
We set the `stack_top` to the start address of the `stack_end` page here: https://github.com/rust-osdev/bootloader/blob/669e3c031f6ee6ad0cbe3f7d09e1b62e8bc34ff1/common/src/lib.rs#L485

However, it looks like `stack_end` is an inclusive bound: https://github.com/rust-osdev/bootloader/blob/669e3c031f6ee6ad0cbe3f7d09e1b62e8bc34ff1/common/src/lib.rs#L183-L186

So I think the stack pointer should be initialized with `(stack_end + 1).start_address()` instead.

Reported by `klim` on the Discord of [Theseus OS](https://www.theseus-os.com/).

phil-opp:
The last page of stack memory was not used before (see #294 ) because we initialized the stack pointer with the start address of the inclusive end page of the stack. This PR fixes this by initializing the stack pointer with the exact configured address, aligned to a 16-byte boundary. To verify that this works, I added a new `min_stack` test that tries booting with a kernel stack size of just 3000 bytes.

We also claimed in the docs that we set up a guard page for the kernel stack, but never actually did this. This PR fixes that issue too by allocating an extra page when creating the virtual kernel stack mapping. We then leave the first page of that allocation unmapped so that a page fault will occur when the kernel overflows its stack.

To ensure that user-configured addresses have the correct alignment, I also added an alignment check to the `mapping_addr` function. There is also a new convenience function for allocating page-aligned virtual regions.

Fixes #294


phil-opp:
cc @inhibitor1217 @tsoutsman

PR #337, opened 2023-01-27
Map BIOS stage-4 at lower address to avoid conflicts with the kernel

### LINKED ISSUE #331: Panic in stage 4 in BIOS boot
The bootloader panics on line 242 with `failed to identity map fame PhysFrame[4kib](20a000): PageAlreadyMapped(PhysFrame[4Kib](0x20a000))`

https://github.com/rust-osdev/bootloader/blob/c73786d894c4f32c12835c4ed1c02b412e3a668d/common/src/lib.rs#L229-L244

I checked the mapping for `0x20b000` and QEMU says its mapped to `0x20b000`

A potental fix would be to just ignore the error with an extra match arm ` Err(MapToError::PageAlreadyMapped(frame)) => {} `

phil-opp:
Parts of stage-4 need to be identity-mapped. This can lead to 'page already mapped' errors when the kernel already occupies the virtual address corresponding to the physical address of the stage-4. For example, a typical virtual start address of the kernel is 2MiB. This was where we placed the stage-4 in physical memory before. This commit avoids that conflict by keeping the stage-4 below the 2MiB boundary.

Fixes #331

### DIFF
--- a/bios/stage-2/src/main.rs
+++ b/bios/stage-2/src/main.rs
@@ -24,9 +24,12 @@ mod vesa;
 /// We use this partition type to store the second bootloader stage;
 const BOOTLOADER_SECOND_STAGE_PARTITION_TYPE: u8 = 0x20;
 
-const STAGE_3_DST: *mut u8 = 0x0010_0000 as *mut u8; // 1MiB (typically 14MiB accessible here)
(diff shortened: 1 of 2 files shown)


phil-opp:
@an-owl Could you try whether this fixes #331 for you?

an-owl:
Thanks that works

PR #341, opened 2023-01-28
Make a link in the documentation clickable

asensio-project:
Hello,

In the config crate there is a link that isn't clickable, it's raw. This PR makes it clickable

Thanks!


phil-opp (approved):
Thanks!

Issue #315, opened 2023-01-05
Frame allocator assumes memory map is in ascending start address order


tsoutsman:
Currently, `LegacyFrameAllocator::allocate_frame_from_descriptor` only ever increases `self.next_frame`. This is problematic, as the memory map descriptors aren't guaranteed to be in ascending start address order (at least for UEFI). If the first descriptor provided by the memory map were also the one with the largest start address, the frame allocator would not allocate from the other descriptors.

Rather than checking whether `self.next_frame < self.start_frame` in `allocate_frame_from_descriptor`, we should set `self.next_frame` to the start address of the descriptor when we find the next suitable descriptor in `allocate_frame`.

phil-opp:
Good catch, thanks a lot for reporting! Your proposed solution sounds good to me. To summarize:

- Add a new `min_frame` field to `LegacyFrameAllocator` and set this field in `new_starting_at`. The `next_frame` field should be initialized to 0.
- Set `self.next_frame` to `core::cmp::max(descriptor.start(), self.min_frame)` when selecting a new descriptor. The assignment should happen just after the `Usable` check here: https://github.com/rust-osdev/bootloader/blob/fffa7ceb317328ed19bc0ebaa0028be152095a05/common/src/legacy_memory_region.rs#L250-L254
- Assert that `self.next_frame >= start_frame` in `allocate_frame_from_descriptor` (since this should always be the case now) instead of the current comparison: https://github.com/rust-osdev/bootloader/blob/fffa7ceb317328ed19bc0ebaa0028be152095a05/common/src/legacy_memory_region.rs#L66-L69

I'd be happy to merge a PR that implements the above.

tsoutsman:
Looking at the code for `construct_memory_map`, that would need to be changed as well. We would need to compare `original` and `memory_map` to see which memory regions have already been consumed by the allocator.

phil-opp:
Oh yes, good point!

Freax13:
Could we just sort the memory map?

JarlEvanson:
I don't think Iterators are sortable, so we would need to be able to collect the iterator, which would have to be able to be done on the stack, since we don't have access to a heap or dynamic allocation other than that. I don't think it's possible

Freax13:
[...]

PR #344, opened 2023-02-16
fix issue #315

JarlEvanson:
This pull request should close https://github.com/rust-osdev/bootloader/issues/315.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -729,6 +729,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "ptr_meta"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


phil-opp:
Thanks a lot for tackling this! I saw that you also opened a PR on the `uefi-rs` repo to implement sorting the memory map: https://github.com/rust-osdev/uefi-rs/pull/662. Given that this PR looks like it's closed to be merged, I wonder if it makes sense to wait for it instead of making our frame allocation more complex. What do you think?

JarlEvanson:
I think that it makes sense to wait, rather than introducing more complexity.

phil-opp:
Ok, thanks!

phil-opp:
I just merged #360 to update the UEFI dependency to v0.20. We should now be able to use the new `sort` function.

PR #360, opened 2023-03-27
Update `uefi` dependency to `v0.20`

kennystrawnmusic:
The old version (0.18.0) of the UEFI crate that bootloader 0.11.3 depends on has code that throws a myriad of future incompatibility/deprecation warnings when compilation is attempted ― deprecation warnings that are highlighted in the rust-lang/rust#107457 tracking issue. I therefore created this pull request that both fixes the version of UEFI depended on (0.20 doesn't have the same problems) and bumps the version of bootloader to 0.11.4 to ensure that this fix gets published to crates.io immediately after it's implemented.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ exclude = ["examples/basic", "examples/test_framework"]
 
 [workspace.package]
 # don't forget to update `workspace.dependencies` below
-version = "0.11.3"
+version = "0.11.4"
 license = "MIT/Apache-2.0"
 repository = "https://github.com/rust-osdev/bootloader"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ exclude = ["examples/basic", "examples/test_framework"]
 
 [workspace.package]
 # don't forget to update `workspace.dependencies` below
-version = "0.11.4"
+version = "0.11.3"
 license = "MIT/Apache-2.0"
 repository = "https://github.com/rust-osdev/bootloader"
 
--- a/tests/runner/src/lib.rs
+++ b/tests/runner/src/lib.rs
@@ -1,6 +1,6 @@
 use bootloader::BootConfig;
 use bootloader::DiskImageBuilder;
-use std::{io::Read, path::Path, process::Command};
+use std::path::Path;
 
 pub fn run_test_kernel(kernel_binary_path: &str) {
     run_test_kernel_internal(kernel_binary_path, None, None)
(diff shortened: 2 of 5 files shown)


uefi/src/main.rs
```diff
@@ -136,7 +136,7 @@ fn main_inner(image: Handle, mut st: SystemTable<Boot>) -> Status {
         }
     );
 
-    let mmap_storage = {
+    let _mmap_storage = {
```
phil-opp, line 139: This is no longer needed because of https://github.com/rust-osdev/uefi-rs/pull/653. We should remove this mapping, not just silence the dead code warning.

Cargo.toml
```diff
@@ -31,7 +31,7 @@ exclude = ["examples/basic", "examples/test_framework"]
 
 [workspace.package]
 # don't forget to update `workspace.dependencies` below
-version = "0.11.3"
+version = "0.11.4"
```
phil-opp, line 34: Please don't bump the version here. You can request a new release in the comments, but the decision when to releases is up to the maintainers. Also, we doing the release should be done in a separate PR anyway because it requires additional steps (e.g. changelog creation).

tests/runner/src/lib.rs
```diff
@@ -2,6 +2,7 @@ use bootloader::BootConfig;
 use bootloader::DiskImageBuilder;
 use std::{io::Read, path::Path, process::Command};
 
+#[allow(dead_code)]
```
phil-opp, line 5: Instead of silencing the warning, we should move this constant into `run_qemu`.

tests/runner/src/lib.rs
```diff
@@ -11,6 +12,8 @@ const QEMU_ARGS: &[&str] = &[
     "none",
     "--no-reboot",
 ];
+
+#[allow(dead_code)]
```
phil-opp, line 16: Instead of silencing the warning, we should move this constant into `run_qemu`.

tests/runner/src/lib.rs
```diff
@@ -98,6 +101,7 @@ pub fn run_test_kernel_on_uefi_pxe(out_tftp_path: &Path) {
     run_qemu(args);
 }
 
+#[allow(dead_code)]
```
phil-opp, line 104: We should use the proper `cfg` flags here instead of just silencing the dead code warning:

```suggestion
#[cfg(any(feature = "uefi", feature = "bios"))]
```

phil-opp (commented):
Thanks for the PR! The update `uefi v0.20` looks good. I left some comments inline. Once they're resolved, we can merge this.

kennystrawnmusic:
> Thanks for the PR! The update `uefi v0.20` looks good. I left some comments inline. Once they're resolved, we can merge this.

Going to do this tonight (Pacific time) — with family at the moment.

kennystrawnmusic:
> Thanks for the PR! The update `uefi v0.20` looks good. I left some comments inline. Once they're resolved, we can merge this.

Alright, just reviewed them. Sorry it took longer than expected; been very busy lately.

phil-opp (approved):
Thanks!

PR #362, opened 2023-04-03
Fixed bug stemming from treating an exclusive range as an inclusive r…

JarlEvanson:
This should fix a rare bug regarding treating an exclusive range as an inclusive range when the zero_end is page aligned.


phil-opp (approved):
Good catch, thanks!

PR #365, opened 2023-04-17
Implemented sorting of uefi memory maps #315

JarlEvanson:
Updated the UEFI specific bootloader to sort the memory map before using it.


phil-opp:
Thanks a lot!

PR #372, opened 2023-05-17
uefi: use new FileSystem abstraction

phip1611:
In the recently released uefi version `0.21`, we merged a neat new FileSystem abstraction. This is my first attempt to use it in any relevant upstream project.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,6 +121,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitflags"
+version = "2.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


phip1611:
By the way, the new API is only an additional feature but not mandatory.

PS: It looks like I have merge rights... I cannot remember that I've ever got them, 😁 @phil-opp 


Screenshot_2023-05-18-21-23-35-55_40deb401b9ffe8e1df2f1cc5ba480b12

phil-opp:
Thanks for the PR! The new file system API looks nice. However, I'm not sure if it's a good idea to set up a global allocator. We generally want to avoid allocations as much as possible because they require special handling when we set up the kernel address space later. My concern is that setting up a global allocator makes it too easy to create additional allocations by accident. Is there any way to use the new file system API without a global allocator?



> PS: It looks like I have merge rights... I cannot remember that I've ever got them

That's because you're also an owner of the `rust-osdev` organization now, which gives you full access to all repos. Please only use this special access if absolutely necessary though.

phip1611:
> However, I'm not sure if it's a good idea to set up a global allocator

Yes, you're right. I get your point.

Issue #378, opened 2023-07-01
Problem building outdated proc-macros2 version on nightly


AarSeBail:
When I try and build my kernel with the latest nightly I am hit with the following error from

```
error: failed to run custom build command for `bootloader v0.11.3`

Caused by:
  process didn't exit successfully: `/home/user/Code/project/target/debug/build/bootloader-5b7a829b00971b70/build-script-build` (exit status: 101)
  --- stderr
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
       Ignored package `bootloader-x86_64-bios-stage-3 v0.11.3` is already installed, use --force to override
  warning: be sure to add `/home/user/Code/project/target/debug/build/bootloader-607e72802a02d67f/out/bin` to your PATH to be able to run the installed binaries
      Blocking waiting for file lock on crate metadata
       Ignored package `bootloader-x86_64-bios-boot-sector v0.11.3` is already installed, use --force to override
  warning: be sure to add `/home/user/Code/project/target/debug/build/bootloader-607e72802a02d67f/out/bin` to your PATH to be able to run the installed binaries
       Ignored package `bootloader-x86_64-bios-stage-2 v0.11.3` is already installed, use --force to override
  warning: be sure to add `/home/user/Code/project/target/debug/build/bootloader-607e72802a02d67f/out/bin` to your PATH to be able to run the installed binaries
      Updating crates.io index
      Blocking waiting for file lock on package cache
      Updating crates.io index
    Installing bootloader-x86_64-uefi v0.11.3
      Blocking waiting for file lock on package cache
      Updating crates.io index
    Installing bootloader-x86_64-bios-stage-4 v0.11.3
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
      Updating crates.io index
      Blocking waiting for file lock on package cache
  warning: package `spin v0.9.5` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
      Blocking waiting for file lock on package cache
  warning: package `spin v0.9.5` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
      Updating crates.io index
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
      Updating crates.io index
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
     Compiling compiler_builtins v0.1.93
[...]

PR #379, opened 2023-07-02
Bump versions to fix build on nightly

AarSeBail:
I've bumped the versions of serde and async-process, and changed gpt to a stable git branch until it recieves a proper version bump. For full details see rust-osdev/bootloader#378. The gpt dependency should be reverted once Quyzi/gpt#91 is resolved, so an issue will need to be opened for that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "anyhow"
-version = "1.0.69"
+version = "1.0.71"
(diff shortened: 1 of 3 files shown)


phil-opp:
Thanks for opening this PR! Unfortunately, we cannot use git dependencies for this project because we want to publish to crates.io. So let's hope that @Quyzi releases a new `gpt` version soon.

AarSeBail:
Alright, in the meantime my fork is usable as a git dependency.

PR #390, opened 2023-08-03
RacyCell<T>: Data race allowed on T

kuzeyardabulut:
Hi,
I found a memory-safety/soundness issue in this crate while scanning Rust code for potential vulnerabilities. This PR contains a fix for the issue.

# Issue Description
`RacyCell<T>` unconditionally implements Sync. This allows users to create data races on `T: !Sync`. Such data races can lead to undefined behavior.

https://github.com/rust-osdev/bootloader/blob/c5e5ed2cb38e4779c5a6ca67b07719a4bfdb1298/bios/common/src/racy_cell.rs#L21

### DIFF
--- a/bios/common/src/racy_cell.rs
+++ b/bios/common/src/racy_cell.rs
@@ -18,4 +18,4 @@ impl<T> RacyCell<T> {
 }
 
 unsafe impl<T> Send for RacyCell<T> where T: Send {}
-unsafe impl<T> Sync for RacyCell<T> {}
+unsafe impl<T: Sync> Sync for RacyCell<T> {}


bjorn3:
`RacyCell` doesn't expose any safe method that allows access to the internal value through a shared reference, so the internal value doesn't need to be `Sync` for `RacyCell` to be sync.

The fact that this PR doesn't break CI suggests that RacyCell isn't needed however. I think all uses can be replaced with the unstable `core::cell::SyncOnceCell` which does have the `T: Sync` bound on it's `Sync` impl.

kuzeyardabulut:
I agree with what you said. This may not cause a direct data race. But making the above changes can be useful.

phil-opp:
> The fact that this PR doesn't break CI suggests that RacyCell isn't needed however. I think all uses can be replaced with the unstable `core::cell::SyncOnceCell` which does have the `T: Sync` bound on it's `Sync` impl.

I guess you mean the `SyncUnsafeCell`? Yes, we could use that as well, but I would prefer to wait until it's stabilized.

phil-opp (approved):
Looks reasonable to me, thanks for submitting!

PR #395, opened 2023-10-17
Embed bios and uefi binaries

mysteriouslyseeing:
This crate embeds links to bios and uefi binaries created in the build.rs script, and that means you can't depend on it with a crate from crates.io, as the binaries created are deleted after the final binary (with the code that you actually wrote) is moved to .cargo/bin/ and it throws a file not found error.
This was kind of annoying to pin down, because when I manually built it, it worked, until it stopped randomly. This was caused (although I didn't know this at the time) by the files being cleaned from the directory they were built in, no matter where the final binary was at that point.

So here is my solution; feature-gated binary embedding using include_bytes!()
I had to rewrite part of mbr but the api is exactly the same when the feature is disabled.

The main reason I wanted to use it as a dependency was to make a command-line interface to replace bootimage, which doesn't work with 10.x or 11.x (the resulting command-line crate is [here](https://crates.io/crates/bootloader_linker)) and I liked the crate structure that bootimage encouraged (main.rs, entry_point!(), lib.rs etc.) as opposed to the crate structure currently recommended (a build.rs that links bootloader to a subpackage containing the kernel and a main.rs that calls qemu).

Anyway, if this is all a terrible idea, please let me know.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,14 +151,8 @@ impl DiskImageBuilder {
         let stage_3 = FileDataSource::Data(BIOS_STAGE_3.to_vec());
         let stage_4 = FileDataSource::Data(BIOS_STAGE_4.to_vec());
         let mut internal_files = BTreeMap::new();
-        internal_files.insert(
-            BIOS_STAGE_3_NAME,
-            stage_3,
-        );
-        internal_files.insert(
-            BIOS_STAGE_4_NAME,
-            stage_4,
-        );
+        internal_files.insert(BIOS_STAGE_3_NAME, stage_3);
+        internal_files.insert(BIOS_STAGE_4_NAME, stage_4);
         let fat_partition = self
             .create_fat_filesystem_image(internal_files)
             .context("failed to create FAT partition")?;
@@ -192,10 +186,7 @@ impl DiskImageBuilder {
         }
 
         let mut internal_files = BTreeMap::new();
-        internal_files.insert(
-            UEFI_BOOT_FILENAME,
-            get_uefi_bootloader(),
-        );
+        internal_files.insert(UEFI_BOOT_FILENAME, get_uefi_bootloader());
         let fat_partition = self
             .create_fat_filesystem_image(internal_files)
             .context("failed to create FAT partition")?;
@@ -240,7 +231,6 @@ impl DiskImageBuilder {
 
         let to = tftp_path.join(UEFI_TFTP_BOOT_FILENAME);
         write_uefi_bootloader(&to)?;
-        
 
         for f in &self.files {
             let to = tftp_path.join(f.0.deref());
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,16 +41,10 @@ const KERNEL_FILE_NAME: &str = "kernel-x86_64";
 const RAMDISK_FILE_NAME: &str = "ramdisk";
 const CONFIG_FILE_NAME: &str = "boot.json";
 
-#[cfg(all(feature = "embedded_binaries", feature = "uefi"))]
 static UEFI_BOOTLOADER: &'static [u8] = include_bytes!(env!("UEFI_BOOTLOADER_PATH"));
-
-#[cfg(all(feature = "embedded_binaries", feature = "bios"))]
 static BIOS_BOOT_SECTOR: &'static [u8] = include_bytes!(env!("BIOS_BOOT_SECTOR_PATH"));
-#[cfg(all(feature = "embedded_binaries", feature = "bios"))]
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -131,16 +143,50 @@ impl DiskImageBuilder {
         Ok(())
     }
 
+    #[cfg(all(feature = "bios", feature = "embedded_binaries"))]
+    /// Create an MBR disk image for booting on BIOS systems.
+    pub fn create_bios_image(&self, image_path: &Path) -> anyhow::Result<()> {
+        const BIOS_STAGE_3_NAME: &str = "boot-stage-3";
```
mysteriouslyseeing, line 149: expensive clone here

src/lib.rs
```diff
@@ -41,6 +41,18 @@ const KERNEL_FILE_NAME: &str = "kernel-x86_64";
...
+static UEFI_BOOTLOADER: &'static [u8] = include_bytes!(env!("UEFI_BOOTLOADER_PATH"));
+
+#[cfg(all(feature = "embedded_binaries", feature = "bios"))]
+static BIOS_BOOT_SECTOR: &'static [u8] = include_bytes!(env!("BIOS_BOOT_SECTOR_PATH"));
+#[cfg(all(feature = "embedded_binaries", feature = "bios"))]
+static BIOS_STAGE_2: &'static [u8] = include_bytes!(env!("BIOS_STAGE_2_PATH"));
+#[cfg(all(feature = "embedded_binaries", feature = "bios"))]
+static BIOS_STAGE_3: &'static [u8] = include_bytes!(env!("BIOS_STAGE_3_PATH"));
+#[cfg(all(feature = "embedded_binaries", feature = "bios"))]
+static BIOS_STAGE_4: &'static [u8] = include_bytes!(env!("BIOS_STAGE_4_PATH"));
```
phil-opp, line 54: Instead of statics, I think we can use consts here. E.g. `const BIOS_BOOT_SECTOR: &'static [u8] = ...`.

src/lib.rs
```diff
@@ -131,16 +143,50 @@ impl DiskImageBuilder {
         Ok(())
     }
 
+    #[cfg(all(feature = "bios", feature = "embedded_binaries"))]
+    /// Create an MBR disk image for booting on BIOS systems.
+    pub fn create_bios_image(&self, image_path: &Path) -> anyhow::Result<()> {
+        const BIOS_STAGE_3_NAME: &str = "boot-stage-3";
+        const BIOS_STAGE_4_NAME: &str = "boot-stage-4";
+        let stage_3 = FileDataSource::Data(BIOS_STAGE_3.to_vec());
```
phil-opp, line 151: The `to_vec` call copies the data to the heap whenever this function is invoked. I think we could avoid this copy by introducing a new `FileDataSource::Bytes(&'static [u8])` variant, which could be used with statics and const data.

phil-opp (commented):
Thanks a lot for the PR! It took me a while to understand your use case and the actual issue, but looking through https://github.com/mysteriouslyseeing/bootloader_linker/ cleared things up. Sorry about the weird crate design, I can imagine that this issue must be very difficult to debug!

I like the idea of having a self-contained library without any external file dependencies at runtime. So I think we can just always do the `include_bytes` without any feature gates. Or do you see any disadvantage in that (apart from a slightly bigger library size)?

mysteriouslyseeing:
I've made the changes you suggested, but it's now failing a check, specifically the Check docs.rs build.

The reason why is that the build.rs script passes in blank PathBufs into the UEFI_BOOTLOADER_PATH (and co) environment variables when running under the docsrs_dummy_build cfg directive.
build.rs > uefi_main (lines 63 to 67):
```rust
#[cfg(not(docsrs_dummy_build))]
let uefi_path = build_uefi_bootloader(&out_dir).await;
// dummy implementation because docsrs builds have no network access
#[cfg(docsrs_dummy_build)]
let uefi_path = PathBuf::new();
```
The fact that the paths don't point to a file didn't matter for the original implementation without the `include_bytes!`, as the crate was only built, and not run, and therefore the path is never resolved.
Now, because Rust tries to resolve the path during the build process with `include_bytes!`, it doesn't work, as the PathBuf is empty and so is resolved to just point at the root, which is /src/ in this case.
The reason this wasn't an issue in previous commits was (I'm pretty sure) the check wasn't running with the embedded_binaries feature enabled.

This can be fixed changing the build script, so that rather than simply doing nothing if docsrs_dummy_build is set, it will instead populate the `out_dir` directory with dummy empty binaries (with std::fs::File::create).

phil-opp (approved):
Thanks a lot for the updates! Looks good!

PR #397, opened 2023-10-22
Add an additional MB of space to the generated FAT partition

### LINKED ISSUE #396: Image builder sporadically throws ENOSPC on macOS even with 72GB of host space available
After having [suddenly added 1,464 lines of code to my kernel](https://github.com/kennystrawnmusic/cryptos/commit/3d738cbec53cb2ee9796b0f2418d768f5a177361) the UEFI version of this will no longer compile:

Screenshot 2023-10-22 at 7 53 38 AM

This has happened before, but figured it was just a transient issue — ah, not anymore. The instant a large change is made, the bootloader suddenly fails to copy the kernel binary onto the image. Why is this?

kennystrawnmusic:
Increasing the image size by an additional megabyte makes the image compile again

Fixes #396

### DIFF
--- a/src/fat.rs
+++ b/src/fat.rs
@@ -26,7 +26,7 @@ pub fn create_fat_filesystem(
         .truncate(true)
         .open(out_fat_path)
         .unwrap();
-    let fat_size_padded_and_rounded = ((needed_size + 1024 * 64 - 1) / MB + 1) * MB;
+    let fat_size_padded_and_rounded = ((needed_size + 1024 * 64 - 1) / MB + 1) * MB + MB;
     fat_file.set_len(fat_size_padded_and_rounded).unwrap();
 
     // choose a file system label


src/fat.rs
```diff
@@ -26,7 +26,7 @@ pub fn create_fat_filesystem(
         .truncate(true)
         .open(out_fat_path)
         .unwrap();
-    let fat_size_padded_and_rounded = ((needed_size + 1024 * 64 - 1) / MB + 1) * MB;
+    let fat_size_padded_and_rounded = ((needed_size + 1024 * 64 - 1) / MB + MB) * MB;
```
Freax13, line 29: Is that increasing the size by a terabyte? Note that the expression in the parentheses is already multiplied by `MB`, so now it ends up being `(... + MB) * MB`, which should be a TB if I'm not mistaken. Replacing the 1 by a 2 is probably a more storage efficient fix 😀

Freax13 (changes requested):

src/fat.rs
kennystrawnmusic: Alright, doing something else then, hold on…

kennystrawnmusic: Alright, I decided to revert it back to what it was before and instead simply append `+ MB` to the end of that long calculation, since adding a megabyte is far more reasonable than multiplying.

Freax13:
Could you remove the first commit?

Freax13:
A revert commit is still a commit, ideally we'd only have a single commit for this pr. See https://stackoverflow.com/a/42522493 on how to remove commits without creating a revert commits.

src/fat.rs
kennystrawnmusic: Finally got this down to one commit but it took `git reset --hard <the first of my previous commits>` and `git pull https://github.com/rust-osdev/bootloader main --allow-unrelated-histories --rebase` to do it.

src/fat.rs
```diff
@@ -26,7 +26,7 @@ pub fn create_fat_filesystem(
         .truncate(true)
         .open(out_fat_path)
         .unwrap();
-    let fat_size_padded_and_rounded = ((needed_size + 1024 * 64 - 1) / MB + 1) * MB;
+    let fat_size_padded_and_rounded = (((needed_size + 1024 * 64 - 1) / MB + 1) * MB) + MB;
```
phil-opp, line 29: To me it's more readable without the extra parentheses:

```suggestion
    let fat_size_padded_and_rounded = ((needed_size + 1024 * 64 - 1) / MB + 1) * MB + MB;
```

phil-opp (approved):
Looks good, thanks!

I updated the PR title to be more descriptive and added the line _Fixes https://github.com/rust-osdev/bootloader/issues/396_ to the PR description to automatically close the issue after merging.

One minor request: Right now the title of your commit is _"Hard reset per @Freax13's request"_, which is not really useful. Could you change it to something more descriptive? An easy way to do this is to run the following command:

```
git commit --amend --message "Add an additional MB of space to the generated FAT partition"
```

src/fat.rs
kennystrawnmusic: Had to use a hard reset and force-push to incorporate both of these changes without also inadvertently turning 1 commit into 2, but done.

phil-opp (approved):
Thanks!

phil-opp:
By the way, I recommend to use `git rebase --interactive` (or `git rebase -i` for short) to squash and reword commits. It's much easier to use than `git reset --hard` in my opinion. For some basic examples, check out e.g. https://www.sitepoint.com/git-interactive-rebase-guide/ .

PR #415, opened 2024-01-28
[v0.9] Fix data layout for `x86_64-bootloader` target

phil-opp:
Fixes build on latest Rust nightly.

See https://github.com/rust-lang/rust/pull/120062


phil-opp:
Released as v0.9.24

PR #421, opened 2024-02-16
[v0.9] Fix data layout for custom targets for LLVM 18

phil-opp:
Fixes build on latest nightly, which was broken by the update to LLVM 18 in https://github.com/rust-lang/rust/pull/120055.


phil-opp:
Published as v0.9.25

PR #422, opened 2024-02-16
[v0.9] Fix map errors during kernel loading

phil-opp:
- Don't error if a kernel page is already mapped to the correct frame
- Fix: unmap temp page again to enable multiple bss-like sections


phil-opp:
Published as v0.9.26

PR #424, opened 2024-02-16
Fix invalid mapping to zero page caused by off-by-one bug

phil-opp:
The `zero_end` bound is exclusive, but we treat the `end_page` as inclusive. So when `zero_end` is page-aligned, we allocate one additional bss page. If this page was already mapped to some other segment, we remap it to a page with random content.

This is the same bug as https://github.com/rust-osdev/bootloader/pull/362.

Fixes https://github.com/phil-opp/blog_os/issues/1297

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- [Fix: unify flags if multiple segments are mapped to same frame with different flags](https://github.com/rust-osdev/bootloader/pull/423)
+- [Fix invalid mapping to zero page caused by off-by-one bug](https://github.com/rust-osdev/bootloader/pull/424)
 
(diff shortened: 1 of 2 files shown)


phil-opp:
Published as v0.9.27 :rocket:

vinc:
Hi @phil-opp, thanks for the recent updates on the v0.9 branch! :rocket: 

There's a little issue with this last one, it should have removed the unused imports `Translate` and `TranslateResult` that were introduced by the previous one (#423) because they are now causing a warning during compilation.

Would you like a quick PR fixing that?

phil-opp:
Ah, I missed that. Yes, a PR against the `v0.9-base` branch would be appreciated!

vinc:
Great, working on it!

PR #437, opened 2024-04-30
Rename `.cargo/config` to `.cargo/config.toml`

phil-opp:
Fixes a deprecation warning.

Also: Replace remaining uses of `cargo xbuild` in CI script and Readme with `-Zbuild-std`.


phil-opp:
Ok, looks like cargo xbuild (which we still use in this branch for some reason) doesn't support `.cargo/config.toml` files. So this PR doesn't work.

PR #446, opened 2024-06-09
Guard the lower 1MB of memory

### LINKED ISSUE #311: UEFI boot should avoid using the lower 1MB of physical RAM

### LINKED ISSUE #445: Ramdisk marked as Usable region in MemoryRegions
https://github.com/rust-osdev/bootloader/blob/e10010e6cb6cbcbdb82ddf29fb59f3a8d27ff2ff/common/src/legacy_memory_region.rs#L163-L171

https://github.com/rust-osdev/bootloader/blob/e10010e6cb6cbcbdb82ddf29fb59f3a8d27ff2ff/common/src/legacy_memory_region.rs#L205-L208

This assumes that the memory region containing the kernel and the region containing the ramdisk are distinct. However if both fall within the same region only the kernel is marked as a bootloader region and the ramdisk is marked as usable.

Wasabi375:
This commit fixes the review changes in #317.
Most of the credit should go to @jasoncouture. I'm not quite sure if and how I can add him as the original author of the commit.
I decided to not rebase his original branch, because I just didn't want to deal with possible conflicts and the changes were so few that just reimplementing them was faster than reworking his changes.

I hope nobody minds me taking over this PR, but since it's stale for over a year I think it's fine :)

The tests are currently failing but as far as I can tell that is due to an unrelated bug that I uncovered with this change(#445)

Closes #311 
Closes #445

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,8 @@ jobs:
       - uses: r7kamura/rust-problem-matchers@v1.1.0
       - name: Run api tests
         run: cargo test -p bootloader_api
+      - name: Run bootloader common tests
+        run: cargo test -p bootloader-x86_64-common
       - name: Run integration tests
         run: cargo test -- --test-threads 1
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,7 @@ jobs:
       - name: Run api tests
         run: cargo test -p bootloader_api
       - name: Run bootloader common tests
+        if: runner.os == 'Linux'
         run: cargo test -p bootloader-x86_64-common
       - name: Run integration tests
         run: cargo test -- --test-threads 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Run api tests
         run: cargo test -p bootloader_api
       - name: Run bootloader common tests
-        if: runner.os == 'Linux'
+        if: runner.arch == 'x86'
         run: cargo test -p bootloader-x86_64-common
       - name: Run integration tests
         run: cargo test -- --test-threads 1
--- a/bios/stage-4/src/main.rs
+++ b/bios/stage-4/src/main.rs
@@ -223,10 +223,7 @@ fn used_memory_slices(info: &BiosInfo) -> impl Iterator<Item = UsedMemorySlice> 
     // uefi/bios common code
     [info.stage_3, info.stage_4, info.config_file]
         .into_iter()
-        .map(|region| UsedMemorySlice {
-            start: region.start,
-            end: region.start + region.len,
-        })
+        .map(|region| UsedMemorySlice::new_from_len(region.start, region.len))
 }
 
 /// Creates page table abstraction types for both the bootloader and kernel page tables.
--- a/bios/stage-4/src/main.rs
+++ b/bios/stage-4/src/main.rs
@@ -224,8 +224,8 @@ fn used_memory_slices(info: &BiosInfo) -> impl Iterator<Item = UsedMemorySlice> 
     [info.stage_3, info.stage_4, info.config_file]
         .into_iter()
         .map(|region| UsedMemorySlice {
-            start: PhysAddr::new(region.start).as_u64(),
-            len: region.len,
+            start: region.start,
+            end: region.start + region.len,
         })
 }
 
--- a/bios/stage-4/src/main.rs
+++ b/bios/stage-4/src/main.rs
@@ -4,7 +4,8 @@
 use crate::memory_descriptor::MemoryRegion;
 use bootloader_api::info::{FrameBufferInfo, PixelFormat};
 use bootloader_boot_config::{BootConfig, LevelFilter};
-use bootloader_x86_64_bios_common::{BiosFramebufferInfo, BiosInfo, E820MemoryRegion};
+use bootloader_x86_64_bios_common::{BiosFramebufferInfo, BiosInfo, E820MemoryRegion, Region};
+use bootloader_x86_64_common::legacy_memory_region::UsedMemorySlice;
 use bootloader_x86_64_common::RawFrameBufferInfo;
 use bootloader_x86_64_common::{
     legacy_memory_region::LegacyFrameAllocator, load_and_switch_to_kernel, Kernel, PageTables,
@@ -55,9 +56,10 @@ pub extern "C" fn _start(info: &mut BiosInfo) -> ! {
     };
     let kernel_size = info.kernel.len;
     let next_free_frame = PhysFrame::containing_address(PhysAddr::new(info.last_used_addr)) + 1;
-    let mut frame_allocator = LegacyFrameAllocator::new_starting_at(
+    let mut frame_allocator = LegacyFrameAllocator::new_with_used_slices(
         next_free_frame,
         memory_map.iter().copied().map(MemoryRegion),
+        used_memory_slices(info),
     );
 
     // We identity-mapped all memory, so the offset between physical and virtual addresses is 0
@@ -214,6 +216,17 @@ fn init_logger(
     );
 
     framebuffer_info
+}
+
+fn used_memory_slices(info: &BiosInfo) -> impl Iterator<Item = UsedMemorySlice> + Clone {
+    // skip kernel and ramdisk because they are handled individually by the
+    // uefi/bios common code
+    [info.stage_3, info.stage_4, info.config_file]
+        .into_iter()
+        .map(|region| UsedMemorySlice {
+            start: PhysAddr::new(region.start).as_u64(),
+            len: region.len,
+        })
 }
 
 /// Creates page table abstraction types for both the bootloader and kernel page tables.
--- a/bios/stage-4/src/main.rs
+++ b/bios/stage-4/src/main.rs
@@ -4,8 +4,7 @@
 use crate::memory_descriptor::MemoryRegion;
 use bootloader_api::info::{FrameBufferInfo, PixelFormat};
 use bootloader_boot_config::{BootConfig, LevelFilter};
-use bootloader_x86_64_bios_common::{BiosFramebufferInfo, BiosInfo, E820MemoryRegion, Region};
-use bootloader_x86_64_common::legacy_memory_region::UsedMemorySlice;
+use bootloader_x86_64_bios_common::{BiosFramebufferInfo, BiosInfo, E820MemoryRegion};
 use bootloader_x86_64_common::RawFrameBufferInfo;
 use bootloader_x86_64_common::{
     legacy_memory_region::LegacyFrameAllocator, load_and_switch_to_kernel, Kernel, PageTables,
@@ -56,10 +55,9 @@ pub extern "C" fn _start(info: &mut BiosInfo) -> ! {
     };
     let kernel_size = info.kernel.len;
     let next_free_frame = PhysFrame::containing_address(PhysAddr::new(info.last_used_addr)) + 1;
-    let mut frame_allocator = LegacyFrameAllocator::new_with_used_slices(
+    let mut frame_allocator = LegacyFrameAllocator::new_starting_at(
         next_free_frame,
         memory_map.iter().copied().map(MemoryRegion),
-        used_memory_slices(info),
     );
 
     // We identity-mapped all memory, so the offset between physical and virtual addresses is 0
@@ -216,14 +214,6 @@ fn init_logger(
     );
 
     framebuffer_info
-}
-
-fn used_memory_slices(info: &BiosInfo) -> impl Iterator<Item = UsedMemorySlice> + Clone {
-    // skip kernel and ramdisk because they are handled individually by the
-    // uefi/bios common code
-    [info.stage_3, info.stage_4, info.config_file]
-        .into_iter()
-        .map(|region| UsedMemorySlice::new_from_len(region.start, region.len))
 }
 
 /// Creates page table abstraction types for both the bootloader and kernel page tables.
--- a/common/src/legacy_memory_region.rs
+++ b/common/src/legacy_memory_region.rs
@@ -18,6 +18,16 @@ pub struct UsedMemorySlice {
     pub end: u64,
 }
 
+impl UsedMemorySlice {
+    /// Creates a new slice
+    pub fn new_from_len(start: u64, len: u64) -> Self {
+        Self {
+            start,
+            end: start + len,
+        }
+    }
+}
+
 /// Abstraction trait for a memory region returned by the UEFI or BIOS firmware.
 pub trait LegacyMemoryRegion: Copy + core::fmt::Debug {
     /// Returns the physical start address of the region.
@@ -208,22 +218,18 @@ where
         ramdisk_slice_len: u64,
         used_slices: S,
     ) -> impl Iterator<Item = UsedMemorySlice> + Clone {
-        BootloaderUsedMemorySliceIter {
-            bootloader: UsedMemorySlice {
+        [
+            UsedMemorySlice {
                 start: min_frame.start_address().as_u64(),
-                // TODO: unit test that this is not an off by 1
-                end: next_free.start_address() - min_frame.start_address(),
+                end: next_free.start_address().as_u64(),
             },
-            kernel: UsedMemorySlice {
-                start: kernel_slice_start.as_u64(),
-                end: kernel_slice_len,
-            },
-            ramdisk: ramdisk_slice_start.map(|start| UsedMemorySlice {
-                start: start.as_u64(),
-                end: ramdisk_slice_len,
-            }),
-            state: KernelRamIterState::Bootloader,
-        }
+            UsedMemorySlice::new_from_len(kernel_slice_start.as_u64(), kernel_slice_len),
+        ]
+        .into_iter()
+        .chain(
+            ramdisk_slice_start
+                .map(|start| UsedMemorySlice::new_from_len(start.as_u64(), ramdisk_slice_len)),
+        )
         .chain(used_slices)
     }
 
@@ -366,43 +372,5 @@ where
         }
 
         None
-    }
-}
-
-#[derive(Clone)]
-struct BootloaderUsedMemorySliceIter {
-    bootloader: UsedMemorySlice,
-    kernel: UsedMemorySlice,
-    ramdisk: Option<UsedMemorySlice>,
-    state: KernelRamIterState,
-}
-
-#[derive(Clone)]
-enum KernelRamIterState {
-    Bootloader,
-    Kernel,
-    Ramdisk,
-    Done,
-}
-
-impl Iterator for BootloaderUsedMemorySliceIter {
-    type Item = UsedMemorySlice;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        match self.state {
-            KernelRamIterState::Bootloader => {
-                self.state = KernelRamIterState::Kernel;
-                Some(self.bootloader)
-            }
-            KernelRamIterState::Kernel => {
-                self.state = KernelRamIterState::Ramdisk;
-                Some(self.kernel)
-            }
-            KernelRamIterState::Ramdisk => {
-                self.state = KernelRamIterState::Done;
-                self.ramdisk
-            }
-            KernelRamIterState::Done => None,
-        }
     }
 }
--- a/common/src/legacy_memory_region.rs
+++ b/common/src/legacy_memory_region.rs
@@ -5,6 +5,7 @@ use core::{
     mem::MaybeUninit,
 };
 use x86_64::{
+    align_down, align_up,
     structures::paging::{FrameAllocator, PhysFrame, Size4KiB},
     PhysAddr,
 };
@@ -232,9 +233,12 @@ where
                 .map(|start| UsedMemorySlice::new_from_len(start.as_u64(), ramdisk_slice_len)),
         )
         .chain(used_slices)
+        .map(|slice| UsedMemorySlice {
+            start: align_down(slice.start, 0x1000),
+            end: align_up(slice.end, 0x1000),
+        })
     }
 
-    // TODO unit test
     fn split_and_add_region<'a, U>(
         mut region: MemoryRegion,
         regions: &mut [MaybeUninit<MemoryRegion>],
@@ -347,7 +351,6 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use bootloader_api::info::MemoryRegionKind;
 
     #[derive(Copy, Clone, Debug)]
     struct TestMemoryRegion {
--- a/common/src/legacy_memory_region.rs
+++ b/common/src/legacy_memory_region.rs
@@ -390,6 +390,33 @@ mod tests {
     }
 
     #[test]
+    fn test_all_regions_frame_alligned() {
+        let regions = create_single_test_region();
+        let mut allocator = LegacyFrameAllocator::new(regions.into_iter());
+        // allocate at least 1 frame
+        allocator.allocate_frame();
+
+        let mut regions = [MaybeUninit::uninit(); 10];
+        let kernel_slice_start = PhysAddr::new(0x50000);
+        let kernel_slice_len = 0x0500;
+        let ramdisk_slice_start = None;
+        let ramdisk_slice_len = 0;
+
+        let kernel_regions = allocator.construct_memory_map(
+            &mut regions,
+            kernel_slice_start,
+            kernel_slice_len,
+            ramdisk_slice_start,
+            ramdisk_slice_len,
+        );
+
+        for region in kernel_regions.iter() {
+            assert!(region.start % 0x1000 == 0);
+            assert!(region.end % 0x1000 == 0);
+        }
+    }
+
+    #[test]
     fn test_kernel_and_ram_in_same_region() {
         let regions = create_single_test_region();
         let mut allocator = LegacyFrameAllocator::new(regions.into_iter());
--- a/common/src/legacy_memory_region.rs
+++ b/common/src/legacy_memory_region.rs
@@ -1,5 +1,6 @@
 use bootloader_api::info::{MemoryRegion, MemoryRegionKind};
 use core::{
+    cmp,
     iter::{empty, Empty},
     mem::MaybeUninit,
 };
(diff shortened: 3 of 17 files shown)


Freax13:
> This commit fixes the review changes in #317. Most of the credit should go to @jasoncouture. I'm not quite sure if and how I can add him as the original author of the commit.

You can add a [Co-authored-by](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) tag to your commit message.

tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
```diff
@@ -0,0 +1,45 @@
...
+        .unwrap();
+        if region.kind == MemoryRegionKind::Usable && region.start < LOWER_MEMORY_END_PAGE {
+            let end = core::cmp::min(region.end, LOWER_MEMORY_END_PAGE);
+            let pages = (end - region.start) / 4096;
+            count += pages;
+        }
+    }
+
+    writeln!(serial(), "Free lower memory page count: {}", count).unwrap();
+    assert!(count > 0x10); // 0x20 chosen arbirarily, we need _some_ free conventional memory, but not all of it. Some, especially on BIOS, may be reserved for hardware.
```
Freax13, line 33:
```suggestion
    assert!(count > 0x10); // 0x10 chosen arbirarily, we need _some_ free conventional memory, but not all of it. Some, especially on BIOS, may be reserved for hardware.
```

tests/test_kernels/lower_memory_free/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "test_kernel_lower_memory_free"
+version = "0.1.0"
+authors = ["Philipp Oppermann <dev@phil-opp.com>"]
```
Freax13, line 4: Please either add yourself as the author or remove the `authors` key.

common/src/legacy_memory_region.rs
```diff
@@ -28,8 +28,12 @@ pub struct LegacyFrameAllocator<I, D> {
     memory_map: I,
     current_descriptor: Option<D>,
     next_frame: PhysFrame,
+    min_frame: PhysFrame,
 }
 
+/// Start address of the first frame that is not part of the lower 1MB of frames
+const LOWER_MEMORY_END_PAGE: u64 = 0x10_000;
```
Freax13, line 35:
```suggestion
/// Start address of the first frame that is not part of the lower 1MB of frames
const LOWER_MEMORY_END_PAGE: u64 = 0x10_0000;
```

common/src/legacy_memory_region.rs
```diff
@@ -40,20 +44,25 @@ where
     /// Skips the frame at physical address zero to avoid potential problems. For example
     /// identity-mapping the frame at address zero is not valid in Rust, because Rust's `core`
     /// library assumes that references can never point to virtual address `0`.  
+    /// Also skips the lower 1MB of frames, there are use cases that require lower conventional memory access (Such as SMP SIPI).
     pub fn new(memory_map: I) -> Self {
         // skip frame 0 because the rust core library does not see 0 as a valid address
-        let start_frame = PhysFrame::containing_address(PhysAddr::new(0x1000));
+        let start_frame = PhysFrame::containing_address(PhysAddr::new(LOWER_MEMORY_END_PAGE));
```
Freax13, line 50: Please update the comment above this line.

Freax13 (commented):

I haven't had time for a full review yet, but here are some preliminary comments.

tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
Wasabi375: nice catch. I guess I should have payed a bit more attention when coping over the existing code

common/src/legacy_memory_region.rs
Wasabi375: Damn. I was sure `0x10_000` was the correct hex representation for 1mb but your right. Not sure how many other places I made the same mistake :frowning:

tests/test_kernels/lower_memory_free/Cargo.toml
Wasabi375: yeah, you're right. I should have changed that. Done

Wasabi375:
> The tests are currently failing but as far as I can tell that is due to an unrelated bug that I uncovered with this change(https://github.com/rust-osdev/bootloader/issues/445)

What should we do about the failing test? Is it ok, to merge this with the tests failing? Should I add the `ignore` attribute to the failing test for now? Or do we wait for a fix to #445?

Freax13:
> > The tests are currently failing but as far as I can tell that is due to an unrelated bug that I uncovered with this change(#445)
> 
> What should we do about the failing test? Is it ok, to merge this with the tests failing? Should I add the `ignore` attribute to the failing test for now? Or do we wait for a fix to #445?

I'd prefer to have #445 fixed first.

Freax13:
> The tests are currently failing but as far as I can tell that is due to an unrelated bug that I uncovered with this change(#445)

The memory map constructed for BIOS boot is wrong. When I apply the following patch, the BIOS test crashes:
```diff
diff --git a/tests/test_kernels/map_phys_mem/src/bin/access_phys_mem.rs b/tests/test_kernels/map_phys_mem/src/bin/access_phys_mem.rs
index 538362f..e0df615 100644
--- a/tests/test_kernels/map_phys_mem/src/bin/access_phys_mem.rs
+++ b/tests/test_kernels/map_phys_mem/src/bin/access_phys_mem.rs
@@ -1,7 +1,7 @@
 #![no_std] // don't link the Rust standard library
 #![no_main] // disable all Rust-level entry points
 
-use bootloader_api::{entry_point, BootInfo};
+use bootloader_api::{entry_point, info::MemoryRegionKind, BootInfo};
 use test_kernel_map_phys_mem::{exit_qemu, QemuExitCode, BOOTLOADER_CONFIG};
 
 entry_point!(kernel_main, config = &BOOTLOADER_CONFIG);
@@ -12,6 +12,18 @@ fn kernel_main(boot_info: &'static mut BootInfo) -> ! {
     let ptr = phys_mem_offset as *const u64;
     let _ = unsafe { *ptr };
 
+    // Write to all usable memory.
+    for region in boot_info
+        .memory_regions
+        .iter()
+        .filter(|region| region.kind == MemoryRegionKind::Usable)
+    {
+        let addr = phys_mem_offset + region.start;
+        let size = region.end - region.start;
+        unsafe {
+            core::ptr::write_bytes(addr as *mut u8, 0xff, size as usize);
+        }
+    }
+
     exit_qemu(QemuExitCode::Success);
 }
 ```

(Note that the `map_phys_mem` test doesn't use a ramdisk, so that can't be the cause of the crash.)

AFAICT this happens because the BIOS code uses `LegacyFrameAllocator::new_starting_at(frame, ...)` with `frame` pointing past the last allocation by the stage-2 loader. Because of this `construct_memory_map` previously marked any memory before  `frame` as being used by the bootloader. With the changes from this PR, this memory is now categorized as `Usable` which is wrong because it also contains all the memory used by the stage-2 loader (e.g. memory for the kernel).

**Comment by Wasabi375** [comment]
I have been thinking about how to best fix this. My current approach is to create a `UsedMemorySlice` abstraction. 
We can then use that to handle marking the regions used by early stages in the bootloader as well as the kernel and ramdisc region. 
The other way is to manually write code for all regions used by the early stages in the bootloader.

The upshot of the new abstraction is that it should automatically fix #445. The downside is that it will probably require me to re-implement `LegacyFrameAllocator::construct_memory_map`. That said the current implementation is quite complicated and the new version should be simpler and hopefully have fewer bugs. 

I'll convert this to a draft PR for now until I have a bit more to show.

**Comment by Wasabi375** [comment]
I was just wondering. Is there a guarantee that memory regions are aligned to frame boundaries or is it ok to return regions that end/start in the middle of a frame?

**Comment by Wasabi375** [comment]
I think I am done with the implementation. It passes all existing tests. 
There is still 1 small bug that is marked with a TODO, but it should be an easy fix. I need to figure out how many regions might be created so that I can allocate enough space. 
I also want to add a few more test cases. I am fairly confident that I didn't add any new bugs, but tests are a good idea anyways. And I guess I should clean up the git commits.

That said I don't think I will have time to continue this week. Maybe at the weekend, so you should not expect any changes until then.

**Comment by Freax13** [comment]
> That said I don't think I will have time to continue this week. Maybe at the weekend, so you should not expect any changes until then.

I'm on vacation this week and won't be able to review the code until the weekend, so there is no rush.

**Comment by Freax13** [comment]
> I was just wondering. Is there a guarantee that memory regions are aligned to frame boundaries or is it ok to return regions that end/start in the middle of a frame?

The UEFI spec requires this (7.2.3 EFI_BOOT_SERVICES.GetMemoryMap()):

image

I think this is true in practice for BIOS/E820 as well.

File: common/src/legacy_memory_region.rs [hunk 5]
```diff
@@ -257,6 +199,125 @@ where
...
+                    end: region.end,
+                    kind: MemoryRegionKind::Usable,
+                };
+                Self::add_region(bootloader, regions, next_index);
+                Self::split_and_add_region(after, regions, next_index, used_slices);
+                return;
+            }
+        }
+        // region is not coverd by any slice
+        Self::add_region(region, regions, next_index);
```
Freax13, line 318:
```suggestion

        // Each loop iteration takes a chunk of `region` and adds it to
        // `regions`. Do this until `region` is empty.
        while region.start != region.end {
            // Check if there is overlap between `region` and `used_slices`.
            if let Some((overlap_start, overlap_end)) = used_slices
                .clone()
                .map(|slice| {
                    // Calculate the start and end points of the overlap
                    // between `slice` and `region`. If `slice` and `region`
                    // don't overlap, the range will be ill-formed
                    // (overlap_start > overlap_end).
                    let overlap_start = cmp::max(region.start, slice.start);
                    let overlap_end = cmp::min(region.end, slice.start + slice.len);
                    (overlap_start, overlap_end)
                })
                .filter(|(overlap_start, overlap_end)| {
                    // Only consider non-empty overlap.
                    overlap_start < overlap_end
                })
                .min_by_key(|&(overlap_start, _)| {
                    // Find the earliest overlap.
                    overlap_start
                })
            {
                // There's no overlapping used slice before `overlap_start`, so
                // we know that memory between `region.start` and
                // `overlap_start` is usable.
                let usable = MemoryRegion {
                    start: region.start,
                    end: overlap_start,
                    kind: MemoryRegionKind::Usable,
                };
                let bootloader = MemoryRegion {
                    start: overlap_start,
                    end: overlap_end,
                    kind: MemoryRegionKind::Bootloader,
                };
                Self::add_region(usable, regions, next_index);
                Self::add_region(bootloader, regions, next_index);

                // Continue after the overlapped region.
                region.start = overlap_end;
            } else {
                // There's no overlap. We can add the whole region.
                Self::add_region(region, regions, next_index);
                break;
            }
        }
```

Here's a much simpler, non-recursive approach.

tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
```diff
@@ -0,0 +1,65 @@
+#![no_std] // don't link the Rust standard library
+#![no_main] // disable all Rust-level entry points
+
+use bootloader_api::{
+    config::Mapping, entry_point, info::MemoryRegionKind, BootInfo, BootloaderConfig,
+};
+use test_kernel_lower_memory_free::{exit_qemu, QemuExitCode};
+
+const LOWER_MEMORY_END_PAGE: u64 = 0x0010_0000;
+const WRITE_TEST_UNTIL: u64 = 0x4000_0000;
```
Freax13, line 10:
```suggestion
const WRITE_TEST_END: u64 = 0x4000_0000;
```

I assume this exists because testing writing to the entire memory is too slow? If so please add a comment stating this.

common/src/legacy_memory_region.rs
```diff
@@ -1,10 +1,23 @@
...
 
+/// A slice of memory that is used by the bootloader and needs to be reserved
+/// in the kernel
+#[derive(Clone, Copy, Debug)]
+pub struct UsedMemorySlice {
+    /// the physical start of the slice
+    pub start: u64,
+    /// the length of the slice
+    pub len: u64,
+}
```
Freax13, line 19: Let's use `start` and `end` instead of `start` and `len`. We never use `len` directly, but only use it to calculate the `end`, so let's just store that instead. We can add a constructor to create a `UsedMemorySlice` from `start` + `len` though.

Wasabi375:
> > I was just wondering. Is there a guarantee that memory regions are aligned to frame boundaries or is it ok to return regions that end/start in the middle of a frame?
> 
> The UEFI spec requires this (7.2.3 EFI_BOOT_SERVICES.GetMemoryMap()):
> 
> image
> 
> I think this is true in practice for BIOS/E820 as well.

So we should also ensure that the memory regions we report to the kernel are page aligned. I don't think this is necessarily true with the current implementation or my changes. I think if the kernel size/ramdisk size is not 4k we break this rule. I guess that is another problem I have to look into for this PR.

common/src/legacy_memory_region.rs
```diff
@@ -257,6 +199,125 @@ where
...
+                start: kernel_slice_start.as_u64(),
+                len: kernel_slice_len,
+            },
+            ramdisk: ramdisk_slice_start.map(|start| UsedMemorySlice {
+                start: start.as_u64(),
+                len: ramdisk_slice_len,
+            }),
+            state: KernelRamIterState::Bootloader,
+        }
+        .chain(used_slices)
```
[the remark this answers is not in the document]
Wasabi375: Oh wow. I spent a good hour on figuring out how to do this. I totally forgot that `Option` implements `IntoIterator`. I always got stuck on the fact that you have 2 different types if you try to call `chain` within an if-statement.

common/src/legacy_memory_region.rs
Wasabi375: Tbh, I prefer the recursive approach but I guess that is a question of style and familiarity. Your version looks like it should also work. Let me add the tests I was planning to write anyway and I can test both approaches. If both work we can use the iterative version. After all I am not a maintainer of this project so your opinion is more important than mine, in regards to style.

Freax13: For me, it's less about recursive vs iterative (though iterative should use less stack memory), and more about the fact that my implementation only has a single if-condition whereas your implementation has 5. My implementation is also significantly shorter (33 non-comment lines vs 69 non-comment lines). My hope (and feel free to say otherwise) is that my implementation is simpler to understand and that it's easier to convince yourself that it's correct.

tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
Wasabi375: I just added this as some kind of defensive programming against slow memory access. But I just tested this and this is pretty much pointless.
1. We don't set a ram size for qemu and the default is 128MB (https://wiki.gentoo.org/wiki/QEMU/Options#RAM)
2. Even if I set this to the maximum on my computer (around 30GB) it only increases the total time of the test from 9.5s to 10.5s
So I will remove this.

Wasabi375: I also moved the write check into its own test, as it is not really related to the lower_memory_free test

common/src/legacy_memory_region.rs
Wasabi375: Yeah maybe your implementation is nicer.

.github/workflows/ci.yml
```diff
@@ -68,6 +68,9 @@ jobs:
       - uses: r7kamura/rust-problem-matchers@v1.1.0
       - name: Run api tests
         run: cargo test -p bootloader_api
+      - name: Run bootloader common tests
+        if: runner.os == 'Linux'
```
Wasabi375, line 72: The `x86_64` crate fails to compile under MacOs. I assume that is because macs use ARM architecture. Not quite sure, what the reason is.

common/src/legacy_memory_region.rs
```diff
@@ -306,3 +343,276 @@ where
...
+        let ramdisk_slice_start = None;
+        let ramdisk_slice_len = 0;
+
+        let kernel_regions = allocator.construct_memory_map(
+            &mut regions,
+            kernel_slice_start,
+            kernel_slice_len,
+            ramdisk_slice_start,
+            ramdisk_slice_len,
+        );
```
Wasabi375, line 411: This test fails, because the kernel slice len is not 4k in size. I assume that is the case with the current implementation as well. I have not yet tested this, but I don't have more time today.

Wasabi375:
> > > I was just wondering. Is there a guarantee that memory regions are aligned to frame boundaries or is it ok to return regions that end/start in the middle of a frame?
> > 
> > 
> > The UEFI spec requires this (7.2.3 EFI_BOOT_SERVICES.GetMemoryMap()):
> > image
> > I think this is true in practice for BIOS/E820 as well.
> 
> So we should also ensure that the memory regions we report to the kernel are page aligned. I don't think this is necessarily true with the current implementation or my changes. I think if the kernel size/ramdisk size is not 4k we break this rule. I guess that is another problem I have to look into for this PR.

I'm not sure this is something we should fix in `construct_memory_map`. I think it would be better if we just assert that this is the case and ensure the page alignment when the kernel or ramdisk, etc is allocated.
Extending the kernel range to be 4k might otherwise overlap the ramdisk range.
Right now this should be fine, I think we always allocate a full frame and just report the smaller size but I'm not sure we should rely on that.

.github/workflows/ci.yml
Freax13: Yup, `x86_64`'s default features include the `instructions` feature, which only compiles on `x86_64`.
```suggestion
        if: runner.arch == 'x64'
```

Freax13:
> I'm not sure this is something we should fix in `construct_memory_map`. I think it would be better if we just assert that this is the case and ensure the page alignment when the kernel or ramdisk, etc is allocated.

While this sounds very reasonable, we also currently report the smaller (i.e. non-aligned) lengths to the kernel, so we'd have to store two lengths internally (one for the real length of whatever was loaded and one for the size of the allocation). 

> Extending the kernel range to be 4k might otherwise overlap the ramdisk range.

Is this a problem? AFAICT we only use these lengths to determine when memory should be marked as unusable and in that case, we don't need byte-level granularity anyway. If we mark slightly more (< 4 KiB) memory as unusable, that should be fine.

> Right now this should be fine, I think we always allocate a full frame and just report the smaller size but I'm not sure we should rely on that.

I think we can just align up the lengths to 4 KiB, can't we?

Wasabi375:
> I think we can just align up the lengths to 4 KiB, can't we?

Yes, but I think that should be done in the bios/uefi specific parts and not at when we generate the regions. That way we can ensure that the regions match the allocations.

Freax13:
> > I think we can just align up the lengths to 4 KiB, can't we?
> 
> Yes, but I think that should be done in the bios/uefi specific parts and not at when we generate the regions. That way we can ensure that the regions match the allocations.

The problem with that is that we also report these lengths to the kernel and we want to tell the kernel the unaligned lengths, so we can't just align the lengths up in the bios/uefi-specific parts.

Wasabi375:
> > > I think we can just align up the lengths to 4 KiB, can't we?
> > 
> > 
> > Yes, but I think that should be done in the bios/uefi specific parts and not at when we generate the regions. That way we can ensure that the regions match the allocations.
> 
> The problem with that is that we also report these lengths to the kernel and we want to tell the kernel the unaligned lengths, so we can't just align the lengths up in the bios/uefi-specific parts.

Yeah, your right. Also I think I overestimated the issues caused if we have overlapping slices we reserve for the bootloader. After all we mark all of them as bootloader regions anyways. Worst case we have one bootloader memory region end in the middle of lets say the kernel region and another bootloader region start right were the first ended. 
Sure it would be nicer if the regions always match the slices exactly, but it does not really matter since the kernel should not use any bootloader region.

In fact we hit this case with the current bios implementation. The config file is not page aligned if the ramdisk length is not a multiple of 4096: 
https://github.com/rust-osdev/bootloader/blob/e10010e6cb6cbcbdb82ddf29fb59f3a8d27ff2ff/bios/stage-2/src/main.rs#L116

I guess this makes it more important that we report the correct sizes to the kernel, otherwise I don't think it has any real impact. In fact it probably saves some memory, because we don't need an entire page for the config file.

Wasabi375:
I think this should now fix all the issues we discussed. Do you want me to cleanup the git commits or are you fine with just merging this as is?

This should close #311 #317  and #445

bios/stage-4/src/main.rs
```diff
@@ -216,6 +218,14 @@ fn init_logger(
     framebuffer_info
 }
 
+fn used_memory_slices(info: &BiosInfo) -> impl Iterator<Item = UsedMemorySlice> + Clone {
+    // skip kernel and ramdisk because they are handled individually by the
+    // uefi/bios common code
+    [info.stage_3, info.stage_4, info.config_file]
```
Freax13, line 224: None of these memory regions are needed after the kernel starts executing -> these don't have to be marked as being used by the bootloader. This also makes the `LegacyFrameAllocator::new_with_used_slices` redundant.

tests/test_kernels/lower_memory_free/src/bin/lower_memory_free.rs
```diff
@@ -0,0 +1,53 @@
...
+        .unwrap();
+        if region.kind == MemoryRegionKind::Usable && region.start < LOWER_MEMORY_END_PAGE {
+            let end = core::cmp::min(region.end, LOWER_MEMORY_END_PAGE);
+            let pages = (end - region.start) / 4096;
+            count += pages;
+        }
+    }
+
+    writeln!(serial(), "Free lower memory page count: {}", count).unwrap();
+    assert!(count > 0x10); // 0x10 chosen arbirarily, we need _some_ free conventional memory, but not all of it. Some, especially on BIOS, may be reserved for hardware.
```
Freax13, line 41:
```suggestion
    assert!(count > 0x10); // 0x10 chosen arbitrarily, we need _some_ free conventional memory, but not all of it. Some, especially on BIOS, may be reserved for hardware.
```

Freax13 (commented):
> I think this should now fix all the issues we discussed.

I added one last comment, but aside from that, this PR looks good to me.

> Do you want me to cleanup the git commits or are you fine with just merging this as is?

Feel free to clean up the comments if you want to, but I'd also be okay with merging this as is.

bios/stage-4/src/main.rs
Wasabi375: I'm not 100% sure about stage 4 and the config file, but we need at least stage 3, because the gdt is part of it's data region. Otherwise the kernel might accidentally overwrite the gdt if it is not changing to a custom gdt.

Wasabi375: I think this is what originally made this fail:
https://github.com/rust-osdev/bootloader/pull/446#issuecomment-2158821382
> > The tests are currently failing but as far as I can tell that is due to an unrelated bug that I uncovered with this change(https://github.com/rust-osdev/bootloader/issues/445)
>
> The memory map constructed for BIOS boot is wrong. When I apply the following patch, the BIOS test crashes:

Freax13: > I'm not 100% sure about stage 4 and the config file, but we need at least stage 3, because the gdt is part of it's data region. Otherwise the kernel might accidentally overwrite the gdt if it is not changing to a custom gdt.

The common code (executed with in stage 4) [creates and loads](https://github.com/rust-osdev/bootloader/blob/3b07272dba12fb877f5679ffe07820ce53edccd8/common/src/gdt.rs#L10) a new GDT.

> I think this is what originally made this fail: [#446 (comment)](https://github.com/rust-osdev/bootloader/pull/446#issuecomment-2158821382)
> 
> > > The tests are currently failing but as far as I can tell that is due to an unrelated bug that I uncovered with this change(#445)
> > 
> > 
> > The memory map constructed for BIOS boot is wrong. When I apply the following patch, the BIOS test crashes:

The root of this failure was that memory allocated by stage-2 (e.g. kernel memory) was marked as usable. 

If I change `used_memory_slices` to return an empty iterator, all the tests still pass, so I'm reasonably confident that it's not needed.

Wasabi375: Maybe you are right. I still feel like overwriting the gdt should lead to issues at some point. I'm still learning a lot about this so I don't really know.
Correct me if I am wrong, but interrupts/exceptions should access the gdt. However I don't think we have any tests with interrupts enabled. 
But lets say we create a test that loads an idt with a trivial interrupt handler that just returns. Then we write all usable memory with 0. And then we trigger an interrupt. This should cause a fault or am I wrong?
I should find some time tomorrow to create said test, but I am pretty confident that this would fail.

Freax13: > Maybe you are right. I still feel like overwriting the gdt should lead to issues at some point.

That shouldn't happen though. The common code allocates memory for the GDT through `LegacyFrameAllocator` and so that memory is marked as unusable. The memory backing the GDT is never advertised as usable and so the user shouldn't overwrite it.

>  I'm still learning a lot about this so I don't really know. Correct me if I am wrong, but interrupts/exceptions should access the gdt.

Yup, you're right, `iretq` always reads from the GDT.

> However I don't think we have any tests with interrupts enabled. But lets say we create a test that loads an idt with a trivial interrupt handler that just returns. Then we write all usable memory with 0. And then we trigger an interrupt. This should cause a fault or am I wrong? I should find some time tomorrow to create said test, but I am pretty confident that this would fail.

This would cause a fault, *if* the GDT was marked as usable, but it's not.

Wasabi375: > That shouldn't happen though. The common code allocates memory for the GDT through LegacyFrameAllocator and so that memory is marked as unusable. The memory backing the GDT is never advertised as usable and so the user shouldn't overwrite it.

Now that you say that. I have been looking at the temporary gdt used for switching to long mode.....
Yeah your right. It should be save to delete this piece of code.

Freax13 (approved):
LGTM!

Thank you for all the work you put in!

Wasabi375:
> Thank you for all the work you put in!

To be fair, I did expect it to be much less work, when I first looked at #317.

That original PR can probably be closed as well.

PR #448, opened 2024-07-11
always cover at least the first 4 GiB of physical memory

### LINKED ISSUE #447: APIC init failed on UEFI, but work on BIOS
I have added ioapic to my code, and keyboard is working.

When I select boot with UEFI, there is a panic on 0x200FEE0320 or something like this, I forget it.

So an error happens when I init Timer IRQ.

Forgive my bad grammar.

What should I do for this?

Freax13:
Always cover at least the first 4 GiB of physical memory. That area contains useful MMIO regions (local APIC, I/O APIC, PCI bars) that we want to make accessible to the kernel even if no DRAM exists >4GiB.

Closes #447
Cc @xuanplus

### DIFF
--- a/common/src/legacy_memory_region.rs
+++ b/common/src/legacy_memory_region.rs
@@ -132,11 +132,17 @@ where
     ///
     /// Useful for creating a mapping for all physical memory.
     pub fn max_phys_addr(&self) -> PhysAddr {
-        self.original
+        let max = self
+            .original
             .clone()
             .map(|r| r.start() + r.len())
             .max()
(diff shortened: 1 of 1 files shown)


Freax13:
CI is failing because semver checks doesn't support the current nightly.

phil-opp (approved):

phil-opp:
Thanks!



> CI is failing because semver checks doesn't support the current nightly.

Maybe we should pin the semver checks job to some older nightly?

PR #453, opened 2024-08-21
Fixed "jc fail" instructions not working properly and updated README.md

spencer3035:
I noticed that the times where "jc fail" asm instructions are called are not behaving as expected. They need to be "call fail" in order for the stack to be set up properly to print the fail character. This can be reproduced by changing the jc to jnc and running in qemu. 

I added a label in the boot.s that just calls the rust "fail" function and it seems to work as expected now. 

I also updated the readme and fixed some error codes that were not unique.

### DIFF
--- a/bios/boot_sector/README.md
+++ b/bios/boot_sector/README.md
@@ -9,7 +9,7 @@ This executable needs to fit into the 512-byte boot sector, so we need to use al
 
 To run in QEMU:
 
-- `qemu-system-x86_64 -drive format=raw,file=../../target/disk_image.bin`
+- `qemu-system-x86_64 -drive format=raw,file=../../target/disk_image.img`
 
 To print the contents of the ELF file, e.g. for trying to bring the size down:
 
--- a/bios/boot_sector/README.md
+++ b/bios/boot_sector/README.md
@@ -4,13 +4,13 @@ This executable needs to fit into the 512-byte boot sector, so we need to use al
 
 ## Build Commands
 
-1. `cargo build --release -Zbuild-std=core --target x86-16bit.json -Zbuild-std-features=compiler-builtins-mem`
-2. `objcopy -I elf32-i386 -O binary target/x86-16bit/release/first_stage target/disk_image.bin
+1. `cargo build --profile=stage-1 -Zbuild-std=core --target ../../i386-code16-boot-sector.json -Zbuild-std-features=compiler-builtins-mem`
+2. `objcopy -I elf32-i386 -O binary ../../target/i386-code16-boot-sector/stage-1/bootloader-x86_64-bios-boot-sector ../../target/disk_image.img`
 
 To run in QEMU:
 
-- `qemu-system-x86_64 -drive format=raw,file=target/disk_image.bin`
+- `qemu-system-x86_64 -drive format=raw,file=../../target/disk_image.bin`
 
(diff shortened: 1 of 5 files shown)


bios/boot_sector/README.md
```diff
@@ -4,13 +4,13 @@ This executable needs to fit into the 512-byte boot sector, so we need to use al
...
 
-1. `cargo build --release -Zbuild-std=core --target x86-16bit.json -Zbuild-std-features=compiler-builtins-mem`
-2. `objcopy -I elf32-i386 -O binary target/x86-16bit/release/first_stage target/disk_image.bin
+1. `cargo build --profile=stage-1 -Zbuild-std=core --target ../../i386-code16-boot-sector.json -Zbuild-std-features=compiler-builtins-mem`
+2. `objcopy -I elf32-i386 -O binary ../../target/i386-code16-boot-sector/stage-1/bootloader-x86_64-bios-boot-sector ../../target/disk_image.img`
 
 To run in QEMU:
 
-- `qemu-system-x86_64 -drive format=raw,file=target/disk_image.bin`
+- `qemu-system-x86_64 -drive format=raw,file=../../target/disk_image.bin`
```
Freax13, line 12:
```suggestion
- `qemu-system-x86_64 -drive format=raw,file=../../target/disk_image.img`
```

bios/boot_sector/src/boot.s
```diff
@@ -36,13 +36,19 @@ check_int13h_extensions:
...
 rust:
     # push arguments
     push dx     # disk number
     call first_stage
+    # Fail code if first stage returns
+    push 'x'
+
+    # Call rust "fail" function that prints the character on the top of the stack
+fail_asm:
+    call fail
```
Freax13, line 51: All other labels in asm don't have a `_asm` suffix, so it seems a bit confusing to me to add one here. How about `call_fail`?

bios/boot_sector/src/dap.rs
```diff
@@ -38,11 +38,11 @@ impl DiskAddressPacket {
         let self_addr = self as *const Self as u16;
         unsafe {
             asm!(
-                "push 0x7a", // error code `z`, passed to `fail` on error
+                "push 'z'", // error code `z`, passed to `fail` on error
```
Freax13, line 41: The same improvement can be done in a couple of other places.

Freax13 (changes requested):

Good catch! The same bug exists in stage-2. Could you fix it there as well?

bios/boot_sector/src/boot.s
spencer3035: I just removed the label jump and added conditional relative jumps to call fail. The additional asm label seemed like a lot of indirection for control flow.

bios/boot_sector/src/dap.rs
spencer3035: I only found this in one place under stage-2, where I made the same change.

bios/boot_sector/README.md
spencer3035: Good catch. If only you could compile check markdown like rust

spencer3035:
I addressed the recommended changes in this new commit. I can squash into one commit if you would like.

Freax13 (approved):
LGTM! Thanks again!

> I can squash into one commit if you would like.

That's alright, I can also do that when merging the PR.

PR #457, opened 2024-09-17
Fix the bootloader hanging during uefi exit_boot_services

### LINKED ISSUE #400: Boot prcess gets sometimes stuck at "TRACE: exiting boot services" when using Uefi and qemu
When running `qemu-system-x86_64 -bios {ovmf_pure_efi_path} -drive format=raw,file={uefi_path} -smp {core_count}`
the bootloader gets sometimes stuck at:
```
INFO : UEFI bootloader started
INFO : Using framebuffer at 0xc0000000
INFO : Reading configuration from disk was successful
INFO : Trying to load ramdisk via Disk
INFO : Ramdisk not found.
TRACE: exiting boot services
```

Higher core counts for the -smp argument result in a higher likelihood of it getting stuck.
(I am not sure if it even occurred once at -smp 1) 

- ovmf-prebuilt = "0.1.0-alpha.1"
- occurred independent of used BootloaderConfig
- never encountered on hardware 
- occurred on Linux, Mac and Windows hosts
- my toy Os repo: https://github.com/daniel-keitel/Steelmind_OS

Wasabi375:
Fixes #400 

Switch from using the "-bios" flag to using "-pflash" instead.
Using "-bios" is not recomended see:
https://lists.katacontainers.io/pipermail/kata-dev/2021-January/001650.html

This also allows us to easily update the ovmf-prebuild binaries.

This fixes https://github.com/rust-osdev/bootloader/issues/400. I'm not sure whether the issue was an outdated ovmf
version or the bios-pflash change. I can't tell what version is
published with the ovmf-prebuild crate, so I can't check.

Ovmf binaries are no longer provided by the ovmf-prebuild crate.
Instead we now download them from the ovmf-prebuild github releases of
the same crate. Unlike the crate, those are actually updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,16 @@
 version = 3
 
 [[package]]
-name = "anyhow"
-version = "1.0.71"
+name = "adler2"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c7d0618f0e0b7e8ff11427422b64564d5fb0be1940354bfe2e0529b18a9d9b8"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
+name = "anyhow"
+version = "1.0.89"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "86fdf8605db99b54d3cd748a44c6d04df638eb5dafb219b135d0149bd0db01f6"
 
 [[package]]
 name = "arrayvec"
@@ -39,7 +45,7 @@ dependencies = [
  "log",
  "parking",
  "polling",
- "rustix",
+ "rustix 0.37.26",
  "slab",
  "socket2",
  "waker-fn",
@@ -67,9 +73,9 @@ dependencies = [
  "cfg-if",
  "event-listener",
  "futures-lite",
- "rustix",
+ "rustix 0.37.26",
  "signal-hook",
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -100,6 +106,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
 name = "bincode"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -122,9 +134,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.3.3"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "630be753d4e58660abd17930c71b647fe46c27ea6b63cc59e1e3851406972e42"
+checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
 
 [[package]]
 name = "bitvec"
@@ -139,6 +151,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
 name = "blocking"
 version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -148,7 +169,7 @@ dependencies = [
  "async-lock",
  "async-task",
  "atomic-waker",
- "fastrand",
+ "fastrand 1.9.0",
  "futures-lite",
  "log",
 ]
@@ -270,9 +291,15 @@ dependencies = [
 name = "bootloader_test_runner"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "bootloader",
- "ovmf-prebuilt",
+ "fs-err",
+ "lzma-rs",
+ "sha2",
  "strip-ansi-escapes",
+ "tar",
+ "tempfile",
+ "ureq",
  "wait-timeout",
 ]
 
@@ -284,9 +311,12 @@ checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
 name = "cc"
-version = "1.0.79"
+version = "1.1.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50d30906286121d95be3d479533b458f87493b30a4b5f79a607db8f5d11aa91f"
+checksum = "2d74707dde2ba56f86ae90effb3b43ddd369504387e718014de010cec7959800"
+dependencies = [
+ "shlex",
+]
 
 [[package]]
 name = "cfg-if"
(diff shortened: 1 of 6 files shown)


Wasabi375:
The `test/runner/src/ovmf.rs` file is based on https://github.com/rust-osdev/uefi-rs/blob/ac19526656953c32e8e0ef7bc703f7700b151ae4/xtask/src/qemu.rs#L65

I am 99% certain that this fixes the issue in #400. However I want to implement the same fix in my own kernel to make sure. 

I also still need to update https://github.com/rust-osdev/bootloader/blob/main/docs/create-disk-image.md

Wasabi375:
Ok, I can confirm that using pflash works on my machine. I don't really understand why the CI tests fails. Maybe someone else can try this fix on their machine and give me feedback. The CI log is not really helpful.

Wasabi375:
I wonder if we should provide most of the code in `test/runner/src/ovmf.rs` as a library. Maybe as an updated version of `ovmf-prebuild`. 
Otherwise this is a lot of code that needs to be copied into any project using the bootloader.

tests/runner/src/lib.rs
```diff
@@ -105,6 +108,8 @@ where
         "-display",
         "none",
         "--no-reboot",
+        "-smp",
+        "8",
```
bjorn3: Blog os at least doesn't support more than 1 core.

tests/runner/src/lib.rs
```diff
@@ -138,10 +143,15 @@ where
         )
     });
 
-    let exit_status = child.wait().unwrap();
+    let Some(exit_status) = child.wait_timeout(Duration::new(120, 0)).unwrap() else {
+        child
+            .kill()
+            .expect("Qemu could not be killed after timeout");
+        panic!("Test timed out after 2 minutes");
+    };
```
bjorn3, line 173: Rust's default test runner never times out. At worst it shows a warning after 60s. A timeout means that if you suspend the testing for longer than 120s using ctrl-z, you get an unnnecessary test failure. Also depending on how complex the tests of your OS are you may genuinely need more than 120s on low-end hardware. On local systems you can always kill the test manually and on CI there should be a timeout from the CI provider anyway.

tests/runner/Cargo.toml
```diff
@@ -9,10 +9,16 @@ edition = "2021"
...
 strip-ansi-escapes = "0.1.1"
-ovmf-prebuilt = { version = "0.1.0-alpha.1", optional = true }
 wait-timeout = "0.2.0"
+anyhow = "1.0.89"
+ureq = "2.10.1"
+fs-err = "2.11.0"
+sha2 = "0.10.8"
+tar = "0.4.41"
+lzma-rs = "0.3.0"
+tempfile = "3.12.0"
```
bjorn3, line 24: Maybe this downloading logic should be added to the ovmf-prebuilt repo instead. There is already a stub crate for this in that repo anyway.

tests/runner/src/omvf.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    /// Creates a copy with a writable, temp copy
+    pub fn with_temp_vars(&mut self) -> Result<()> {
+        let temp_dir = TempDir::new()?;
+        let temp_path = temp_dir.path();
+
+        // Make a copy of the OVMF vars file so that it can be used
+        // read+write without modifying the original. Under AArch64, some
+        // versions of OVMF won't boot if the vars file isn't writeable.
```
bjorn3, line 143: Maybe reading the file and writing it to a new file would work to avoid an explicit permission change?

tests/runner/Cargo.toml
```diff
@@ -9,10 +9,16 @@ edition = "2021"
...
+uefi = ["bootloader/uefi"]
 
 [dependencies]
 bootloader = { path = "../..", default-features = false }
 strip-ansi-escapes = "0.1.1"
-ovmf-prebuilt = { version = "0.1.0-alpha.1", optional = true }
 wait-timeout = "0.2.0"
+anyhow = "1.0.89"
+ureq = "2.10.1"
+fs-err = "2.11.0"
```
bjorn3, line 20: It seems easy to avoid this extra dependency by writing out the error handling code yourself. It is only used in a couple of places anyway.

tests/runner/src/omvf.rs
```diff
@@ -0,0 +1,261 @@
...
+    path::{Path, PathBuf},
+};
+use tar::Archive;
+use tempfile::TempDir;
+use ureq::Agent;
+#[cfg(target_os = "linux")]
+use {std::fs::Permissions, std::os::unix::fs::PermissionsExt};
+
+/// Name of the ovmf-prebuilt release tag.
+const OVMF_PREBUILT_TAG: &str = "edk2-stable202311-r2";
```
bjorn3, line 15: This is a pretty outdated version, right?

tests/runner/Cargo.toml
Freax13: Related: https://github.com/rust-osdev/ovmf-prebuilt/issues/38, https://github.com/nicholasbishop/ovmf-prebuilt/commit/73d349224b2f6ab092eb4c736db2aa0e9bed46f3.

tests/runner/src/lib.rs
Wasabi375: Not sure why that matters here. This is just for the tests in this repo and has no impact on any kernel build with this bootloader. Also correct me if I am wrong but Blog Os should work fine on a multicore system. It will use just 1, but the other cores don't do anything unless specifically started by the kernel.

tests/runner/src/lib.rs
Wasabi375: I don't mind removing this. I don't think the crtl-z case is common, but maybe that just comes down to preference. I never use that myself. 
I mostly added this so I could easily run the tests locally in the background and get notified about failures. 
This bug is random in nature so I ran the tests locally in a loop so I needed some kind of failure state.

As for CI the timeout is 30minutes and that seems excessive. Especially when like here the CPU is maxed out by the hanging test.

> Also depending on how complex the tests of your OS are you may genuinely need more than 120s on low-end hardware.

That might actually explain the failing CI tests. So maybe I should remove this.

tests/runner/Cargo.toml
Wasabi375: I agree. Maybe that should be part of a separate PR though.

tests/runner/src/omvf.rs
Wasabi375: I'm not sure what you mean. I am creating a copy of the original file. Or are you referring to this?
```
// Necessary, as for example on NixOS, the files are read-only inside
// the Nix store.
#[cfg(target_os = "linux")]
fs_err::set_permissions(&ovmf_vars, Permissions::from_mode(0o666))?;
```
I copied that from the uefi crates tests. I did not really look into this too much. I assume they had some issues that meant they needed to add this.

tests/runner/Cargo.toml
Wasabi375: Your right fs-err and anyhow are not really necessary here.

tests/runner/src/omvf.rs
Wasabi375: Yes. It's about a year old. I chose this version as it is the one the uefi crate uses in their tests. 
We can probably update this. It just wasn't a priority.

tests/runner/src/omvf.rs
bjorn3: If you use fs::copy you make a copy of the file including pernissions and xattrs. It seems that here a copy of just the file contents ignoring permissions would be a better option.

Wasabi375: I'm not convinced that would be any cleaner though. A simple call to `set_permissions` is cleaner in my opinion. 
Also this way we get some possible advantages from COW file systems, etc.

Wasabi375:
Do you know how I can get the actual logs from qemu in the CI ? Right now the tests time after 30 minutes but I can't reproduce this locally.

tests/runner/src/omvf.rs
Wasabi375: I just tested the latest tarball version(`edk2-stable202405-r1`) released to https://github.com/rust-osdev/ovmf-prebuilt.

It breaks the uefi pxe tests.  I don't really know enough about pxe to debug this, but it looks like it does not find the bootloader and tries to start a uefi terminal.

Cargo.lock
```diff
@@ -1310,13 +1679,29 @@ version = "0.48.1"
...
+ "windows_i686_gnu 0.48.0",
+ "windows_i686_msvc 0.48.0",
+ "windows_x86_64_gnu 0.48.0",
+ "windows_x86_64_gnullvm 0.48.0",
+ "windows_x86_64_msvc 0.48.0",
+]
+
+[[package]]
+name = "windows-targets"
+version = "0.52.6"
```
bjorn3, line 1693: This is now pulling in an extra copy of windows-targets. Maybe update the async-process dependency of bootloader to 2.3.0, which uses windows-targets 0.52.6 too?

tests/runner/src/omvf.rs
bjorn3: At the very least make it `#[cfg(unix)]` then.

tests/runner/Cargo.toml
nicholasbishop: FYI this is available now: https://crates.io/crates/ovmf-prebuilt/0.2.1

PR #460, opened 2024-10-07
Added some docs and an example for how to use the APIC

Voinea-Radu:
I do believe that this could be useful when migrating from version 0.9 to version 0.11, as well as a quick way to get things going if starting from nothing.


Freax13:

I'm not sure if we should adopt this example: This example goes way beyond what's needed to get started in setting up a small kernel with the bootloader. I'm not sure if we want this repo to serve as a place for various examples of next-steps one could take in kernel dev.

Perhaps this is better suited for the [x2apic](https://github.com/kwzhao/x2apic-rs) crate. Maybe it could live in its own repo entirely.

Voinea-Radu:
Hi there, thank you a lot for taking the time to look into this
I do understand that it might be a bit complex for a small kernel, but these were what I needed to port the 0.9 blog version to the 0.11 current version and would have liked to have had them just be available somewhere so I can read thru them. I do find that especially some of the structs that hold all the values are unnecessary for a small startup and would be more than happy to remove them along side any other _future proofing_

docs/apic_example/apic.md
ChocolateLoverRaj, line 1: It would be nicer to look at on GitHub if this was called `README.md`

ChocolateLoverRaj (changes requested):

ChocolateLoverRaj:
> I'm not sure if we want this repo to serve as a place for various examples of next-steps one could take in kernel dev.

It doesn't have to be in this repo, but it is actually a very cool idea to have various examples of next steps. It would be very helpful. There are many people who are following https://os.phil-opp.com/, are done reading edition-2, are done reading edition-3 so far, and are figuring out how to do things rather than waiting for more stuff in edition-3. It would be very helpful if there was a place where people could share examples with each other.

SauravMaheshkar:
> > I'm not sure if we want this repo to serve as a place for various examples of next-steps one could take in kernel dev.
> 
> It doesn't have to be in this repo, but it is actually a very cool idea to have various examples of next steps. It would be very helpful. There are many people who are following https://os.phil-opp.com/, are done reading edition-2, are done reading edition-3 so far, and are figuring out how to do things rather than waiting for more stuff in edition-3. It would be very helpful if there was a place where people could share examples with each other.

Agree about various examples! It'd be nice if we could curate a list of other projects based on https://os.phil-opp.com/ that go beyond the blogs. Maybe a simple discussion thread on the original repository https://github.com/phil-opp/blog_os/discussions ?

SauravMaheshkar:
PS: Would love to see a more minimal version of this in any way be it this repository, the `blog_os` project or anywhere else. This was really helpful to me. Thanks @Radu-Voinea

PR #461, opened 2024-10-12
mention E820 in docs for UnknownBios

### LINKED ISSUE #459: What does the value in MemoryRegionKind::UnknownBios(u32) mean?
What does the number contained in it mean? Is this the e820 memory type? I think it's worth documenting.

Freax13:
This wasn't entirely obvious previously, let's be explicit.

Fixes #459


phil-opp (approved):
Thanks!

PR #464, opened 2024-10-25
Copy all old pages table entries

### LINKED ISSUE #462: Hang after loading first level 4 table on uefi

Colepng:
Corrects the assumption that only the first entry is the active page table is used.

closes #462

### DIFF
--- a/uefi/src/main.rs
+++ b/uefi/src/main.rs
@@ -410,9 +410,10 @@ fn create_page_tables(
             }
         };
 
-        // copy the first entry (we don't need to access more than 512 GiB; also, some UEFI
-        // implementations seem to create an level 4 table entry 0 in all slots)
-        new_table[0] = old_table[0].clone();
+        // copy all entries
+        for (index, entry) in old_table.iter().enumerate() {
+            new_table[index] = entry.clone();
+        }
 
         // the first level 4 table entry is now identical, so we can just load the new one
         unsafe {


Freax13:
Does this panic you saw in #462 still happen with these changes?

Colepng:
Yeah it still panics when trying to allocate boot info. I will try and do some more digging when I have a chance. It's panicking [here](https://github.com/rust-osdev/bootloader/blob/main/common/src/lib.rs#L519-L526).

Image.

Freax13:
Hmm, so the problem is that up until now we've assumed that all existing memory is only in the first pml4 entry and so when we map in more memory, we assume that all other memory is unused:
https://github.com/rust-osdev/bootloader/blob/ea3f61a57ccd57b7ed37eeed1e53ee6cb2cfdd81/common/src/level_4_entries.rs#L42
Obviously, with the changes introduced in this PR, this is no longer the case as, for example, the frame buffer would also be in the address range mapped by the second pml4 entry. I'm looking into a better way to fix this.

Colepng:
Hmm. An simple fix would be to check what entry the framebuffer is in and mark it as used. I can try this patch out but any other devices that are mapped would also cause a similar issue.

Colepng:
Wait my bad there seems to already be a check for the framebuffer address.

Colepng:
So I got it working. All I added was cloning the second entry in the level 4 and marking it used. I don't know how to go about writing a general solution but I would love to help.

Freax13:
> Hmm. An simple fix would be to check what entry the framebuffer is in and mark it as used. I can try this patch out but any other devices that are mapped would also cause a similar issue.
>
> Wait my bad there seems to already be a check for the framebuffer address.

I assume you mean [this check](https://github.com/rust-osdev/bootloader/blob/ea3f61a57ccd57b7ed37eeed1e53ee6cb2cfdd81/common/src/level_4_entries.rs#L72-L76), right? That check operates on the address where the frame buffer will be mapped into the kernel which is likely not the same address where it's mapped in the bootloader.

> So I got it working. All I added was cloning the second entry in the level 4 and marking it used. I don't know how to go about writing a general solution but I would love to help.

We could mark all entries covered by the frame buffer as used and copy them over when creating the new PML4, but I think the cleaner solution is to just never modify the page tables used by the bootloader. Still, I opened a PR based on your idea as well in #466. The big upside is that it's much simpler than my original implementation in #465.

Freax13:
Superseeded by #466

PR #466, opened 2024-10-26
copy more PML4 entries

### LINKED ISSUE #462: Hang after loading first level 4 table on uefi

Freax13:
If there is more than 512 GiB of memory, we need to copy more than the first PML4 entry to access it. Similarly if the frame buffer is not in the address range covered by the first PML4 entry, we need to copy more entries in order to access it. We also need to mark these PML4 entries as unusable when allocating virtual addresses for the kernel because the bootloader maps some of them into its own address space.

This is a much simpler solution to #462 than #465. The downside is that we mark some of the kernel's virtual address space as unusable when it doesn't technically need to be.

@colepng Could you please also test this PR?

Supersedes #464
Closes #462

### DIFF
--- a/common/src/level_4_entries.rs
+++ b/common/src/level_4_entries.rs
@@ -39,7 +39,17 @@ impl UsedLevel4Entries {
             rng: config.mappings.aslr.then(entropy::build_rng),
         };
 
-        used.entry_state[0] = true; // TODO: Can we do this dynamically?
+        // The bootloader maps of the kernel's memory into its own page tables.
+        // We need to prevent overlaps, so mark all memory that could already
+        // be used by the bootload as inaccessible.
+
+        // All memory in this range is identity mapped.
+        used.mark_range_as_used(0, max_phys_addr.as_u64());
+
+        // The bootload needs to access the frame buffer.
+        if let Some(frame_buffer) = framebuffer {
+            used.mark_range_as_used(frame_buffer.addr.as_u64(), frame_buffer.info.byte_len);
+        }
 
         // Mark the statically configured ranges from the config as used.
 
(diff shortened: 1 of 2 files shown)


phil-opp (approved):
Looks good to me, thanks! And sorry again for the long delay!

phil-opp:
> The downside is that we mark some of the kernel's virtual address space as unusable when it doesn't technically need to be.

Just to clarify: Do you mean unusable for the bootloader or unusable in the memory map (i.e. affecting the kernel)?

Freax13:
> Looks good to me, thanks! And sorry again for the long delay!

Thank you for the review!

> > The downside is that we mark some of the kernel's virtual address space as unusable when it doesn't technically need to be.
> 
> Just to clarify: Do you mean unusable for the bootloader or unusable in the memory map (i.e. affecting the kernel)?

We mark it as unusable for dynamic virtual memory space allocations for the kernel i.e. for mappings configured with [`Mapping::Dynamic`](https://docs.rs/bootloader_api/latest/bootloader_api/config/enum.Mapping.html#variant.Dynamic). IIRC mappings with statically configured addresses are not affected because we never do any checks for those.

phil-opp:
Ah right. So when the UEFI firmware puts the framebuffer somewhere into the second PML4 entry, we won't use any address within that PML4 entry for dynamic kernel mappings?

Freax13:
> Ah right. So when the UEFI firmware puts the framebuffer somewhere into the second PML4 entry, we won't use any address within that PML4 entry for dynamic kernel mappings?

Yes, exactly.

Freax13:
> Ok, thanks for clarifying! Personally, I'm fine with this limitation for now. We can then try to make the used-tracking more precise in some future PR.
> 
> I'm also looking into #465 currently. Which of the two approaches do you prefer?

I like how #465 solves this problem without any limitations for the loaded kernel, but I think the code is much uglier/hackier. I agree that losing a pml4 is acceptable for now. I like your idea of making the used tracking more precise, this would allow us to use the less-hacky from this PR code but also minimize wasting virtual address space.

phil-opp:
Ok, then let's proceed with this PR.

I don't mind the indirect memory access in #465, but I agree that the code seems a bit hacky in places (e.g. a separate translation step for each individual byte or the construction of an invalid `&mut` reference).

PR #471, opened 2024-11-01
Remove 3dnow features from stage4 target

Ollrogge:
Remove the 3dnow features from the stage4 target since rust [removed](https://github.com/rust-lang/rust/pull/127864) support for them.

This silences the following messages while building:

```bash
'-3dnow' is not a recognized feature for this target (ignoring feature)
'-3dnowa' is not a recognized feature for this target (ignoring feature)
```


Freax13 (approved):

Looks good to me! :shipit:

PR #474, opened 2024-11-17
Convert LF to CRLF when writing to serial port

### LINKED ISSUE #473: QEMU `serial0` logs look bad because logger doesn't use `\r\n`
This is how QEMU looks like after enabling serial logging with `LockedLogger`:
Image

When using `-serial stdio`, it looks normal.

ChocolateLoverRaj:
Fixes #473 

I'm not sure if this is the best way to implement this in terms of performance.

~~This fixes the issue with `LockedLogger`, but the logging that gets done before that still doesn't use CRLF, which is why I marked this as a draft PR.~~ UPDATE: I forgot to change the `bootloader` dependency in Cargo.toml, after changing it all logging was as expected.

Without this change
image

With this change:
image

### DIFF
--- a/common/src/serial.rs
+++ b/common/src/serial.rs
@@ -17,7 +17,12 @@ impl SerialPort {
 
 impl fmt::Write for SerialPort {
     fn write_str(&mut self, s: &str) -> fmt::Result {
-        self.port.write_str(s).unwrap();
+        for char in s.bytes() {
+            match char {
+                b'\n' => self.port.write_str("\r\n").unwrap(),
+                byte => self.port.send(byte),
+            }
+        }
         Ok(())
     }
 }


common/src/serial.rs
```diff
@@ -17,7 +17,12 @@ impl SerialPort {
 
 impl fmt::Write for SerialPort {
     fn write_str(&mut self, s: &str) -> fmt::Result {
-        self.port.write_str(s).unwrap();
+        for char in s.bytes() {
+            match char {
+                b'\n' => self.port.write_str("\r\n").unwrap(),
+                byte => self.port.send(byte),
+            }
+        }
```
Freax13, line 25: Another fix for #473 might be to change the `writeln!` call to a `write("...\r\n")` call, this wouldn't work if the log message contains multi-line strings, so I think this solution is preferable. 

Performance could certainly be improved a bit, but this code is not at all performance-critical, so let's keep things simple.

Freax13 (approved):

This looks good to me! :shipit:

PR #476, opened 2024-11-23
Remove "UEFI boot" log message

### LINKED ISSUE #475: `log::info!` called before initializing logger
https://github.com/rust-osdev/bootloader/blob/af99f0f9e913cb9741d47de457b942f706f79625/uefi/src/main.rs#L534

This line doesn't do anything write now. It basically functions as a comment. It should either be
- Removed
- Moved to be after calling `init_logger` (I don't think it's necessary since https://github.com/rust-osdev/bootloader/blob/af99f0f9e913cb9741d47de457b942f706f79625/uefi/src/main.rs#L115
- Turned into a comment

ChocolateLoverRaj:
Fixes #475


Freax13 (approved):

LGTM! :shipit:

I rebased this PR onto the latest main branch to fix the issues caused by the latest nightly.

PR #478, opened 2024-11-30
Update x86_64

Makonede:


Freax13:
We just released a new version for x86_64 that fixes the builds on the latest nightlies. Instead of upgrading to a new major version, can you try running `cargo update -p x86_64`?

Makonede:
does this work?

Freax13 (approved):

> does this work?

Yes :partying_face:

Issue #483, opened 2025-01-17
Dependency `instant` is unmaintained


gaojiaqi7:
```
cargo deny check advisories
2025-01-17 02:39:06 [WARN] unable to find a config path, falling back to default config
error[unmaintained]: `instant` is unmaintained
   ┌─ /home/env/source/bootloader/Cargo.lock:60:1
   │
60 │ instant 0.1.13 registry+https://github.com/rust-lang/crates.io-index
   │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
   │
   ├ ID: RUSTSEC-2024-0384
   ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0384
   ├ This crate is no longer maintained, and the author recommends using the maintained [`web-time`] crate instead.
     
     [`web-time`]: https://crates.io/crates/web-time
   ├ Solution: No safe upgrade is available!
   ├ instant v0.1.13
     └── fastrand v1.9.0
         └── futures-lite v1.13.0
             ├── async-io v1.13.0
             │   └── async-process v1.8.1
             │       └── (build) bootloader v0.11.9
             │           └── bootloader_test_runner v0.1.0
             │               └── (dev) bootloader v0.11.9 (*)
             ├── async-process v1.8.1 (*)
             └── futures-concurrency v7.6.2
                 └── (build) bootloader v0.11.9 (*)
```

bjorn3:
Updating futures-lite to 2.6 would remove the dependency on instant. Unfortunately futures-concurrency doesn't have a version which depends on futures-lite 2.6. Only async-process has an updated version available.

gaojiaqi7:
Thanks @bjorn3 @Freax13 ! The `instant` crate is still present in the dependency tree because it is indirectly referenced by `tempfile`:

```
error[unmaintained]: `instant` is unmaintained
   ┌─ /home/env/source/bootloader/Cargo.lock:26:1
   │
26 │ instant 0.1.12 registry+https://github.com/rust-lang/crates.io-index
   │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
   │
   ├ ID: RUSTSEC-2024-0384
[...]

PR #485, opened 2025-01-20
Cargo.lock: update `tempfile`

gaojiaqi7:
Run `cargo update -p tempfile` to bump `tempfile` and remove unmaintained dependency `instant`.

Fixes https://github.com/rust-osdev/bootloader/issues/483

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,12 +264,9 @@ dependencies = [
 
 [[package]]
 name = "fastrand"
-version = "1.9.0"
+version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e51093e27b0797c359783294ca4f0a911c270184cb10f85783b118614a1501be"
-dependencies = [
- "instant",
-]
+checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
 
 [[package]]
 name = "fatfs"
@@ -334,32 +331,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "hermit-abi"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "443144c8cdadd93ebf52ddb4056d257f5b52c04d3c804e657d19eb73fc33668b"
-
-[[package]]
-name = "instant"
-version = "0.1.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a5bbe824c507c5da5956355e86a746d82e0e1464f65d862cc5e71da70e94b2c"
-dependencies = [
- "cfg-if",
-]
-
-[[package]]
-name = "io-lifetimes"
-version = "1.0.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eae7b9aee968036d54dce06cebaefd919e4472e753296daccd6d344e3e2df0c2"
-dependencies = [
- "hermit-abi",
- "libc",
- "windows-sys",
-]
-
-[[package]]
 name = "itoa"
 version = "1.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


phil-opp (approved):
Thanks!

phil-opp:
Looks like the bitflags version bump leads to a dependency on std (see CI errors). We probably need to disable some default features or enable a no_std feature.

Freax13:
Updating tempfile also requires updating rustix. The old rustix version didn't depend on bitflags 2.0, but the new one does. rustix being a std crate, it enables the std feature in bitflags 2.0.
Cargo feature unification makes it so that the features of all dependencies in the whole workspace are unified and so when cargo goes to build our non-std packages, it thinks it needs to build bitflags with std even though that's only required by another package.
There's an accepted RFC to disable workspace-wide feature resolution, but AFAICT. it hasn't been implemented yet. One can also use [cargo-hack](https://github.com/taiki-e/cargo-hack) to work around this.

The only way to fix this I can think of is to remove the tempfile dependency altogether. We could either continue writing to a file and just create the file ourselves or do the processing in-memory. I have patches for the latter, but I'm a little concerned that some users might want to build large-ish images and they might not like the increased memory usage.

Another thing to keep in mind is that the `instant` dependency is very likely not even used in practice. We only depend on it because fastrand 1.9.0 depends on it, but even fastrand only uses it when it's being built for [non-WASI WebAssembly targets](https://docs.rs/crate/fastrand/1.9.0/source/Cargo.toml#23) and I seriously doubt that anything is trying to do that.

gaojiaqi7:
Yes, it looks like the dependencies updated in this PR have introduced new problems. I am going to close this PR because it does not completely solve the issue.

phil-opp:
I think the feature unification only occurs because we run `cargo check -all`, which builds all packages in the workspace for the host target, including the test kernels. Cargo doesn't unify dependency features across different targets, but it does in this case because we build the test kernels for the host target too.

We don't really need to run a separate check command for the test kernels because the they are already built as artifact dependencies in the `dev-dependencies` of the main `bootloader` package. So we can run `cargo check --all-targets --workspace --exclude "test_kernel_*"` instead and this will still build the test kernels, but this time only for their intended target. This should resolve the feature unification issue because cargo doesn't unify features when builting for different targets.

I'll open a PR for this.

Freax13:
> I think the feature unification only occurs because we run `cargo check -all`, which builds all packages in the workspace for the host target, including the test kernels. Cargo doesn't unify dependency features across different targets, but it does in this case because we build the test kernels for the host target too.

Yes, that's correct.

> We don't really need to run a separate check command for the test kernels because the they are already built as artifact dependencies in the `dev-dependencies` of the main `bootloader` package. So we can run `cargo check --all-targets --workspace --exclude "test_kernel_*"` instead and this will still build the test kernels, but this time only for their intended target. This should resolve the feature unification issue because cargo doesn't unify features when builting for different targets.

This should work, but keep in mind that rust-analyzer also uses --all, so if we break --all, this will impact the LSP experience for some (most?) users.

phil-opp:
> This should work, but keep in mind that rust-analyzer also uses --all, so if we break --all, this will impact the LSP experience for some (most?) users.

Hmm, good point. Only when working on this repo or even if the bootloader crate is used as a dependency?

Freax13:
Only the former.

phil-opp:
There is no way to configure rust-analyzer independently of the editior, right? Only e.g. VSCode-specific config.

Freax13:
> There is no way to configure rust-analyzer independently of the editior, right? Only e.g. VSCode-specific config.

I don't know, I only use VSCode.

phil-opp:
I guess the other option could be to make the `test_kernels` dir its own cargo workspace. Perhaps that's the cleaner solution? Then we should not need to configure rust-analyzer.

phil-opp:
> I don't know, I only use VSCode.

Me too. I would be also fine with committing a VSCode-specific config file to the repo that sets the `--exclude "test_kernels_*"` argument for `rust-analyzer`.

Freax13:
> I guess the other option could be to make the `test_kernels` dir its own cargo workspace. Perhaps that's the cleaner solution? Then we should not need to configure rust-analyzer. 

That's the route I took for https://github.com/freax13/mushroom. Other than having to manage multiple lock files, this works well. I think that rust-analyzer can cope quite well with this, but then again, I might have been using this setup for so long that I've just learned to blend out the rough edges.

Freax13:
BTW, here's the RFC that will eventually fix all of this: https://github.com/rust-lang/rfcs/pull/3692

PR #486, opened 2025-01-21
Move test kernels to a separate workspace

phil-opp:
The test kernels need to be built for a different target and we don't want to unify dependency features with the main packages (as this can lead to enabled `std` features).

The test kernels are still built as part of the main `bootloader` package because they are specified as artifact dependencies. For completeness, this PR also adds new check/fmt/clippy steps for the new workspace. We also set a default target for the `tests/test_kernels` dir.

This PR should resolve the issues we saw in https://github.com/rust-osdev/bootloader/pull/485.


Freax13 (approved):
LGTM, thanks! :shipit:

PR #487, opened 2025-01-21
fix condition for running bootloader common tests

Freax13:
There's no 'x86' arch. There is 'X86', but we currently don't any CI job on a 32-bit target, so this still wouldn't work. Let's just use X64.


phil-opp (approved):
Thanks!

PR #491, opened 2025-02-04
change rustc-abi in custom targets to x86-softfloat

Freax13:
With the latest nightly, setting "+soft-float" in "features" is only allowed if "rustc-abi" is set to "x86-softfloat".

Fixes a nightly breakage introduced by https://github.com/rust-lang/rust/pull/136146

### DIFF
--- a/i686-stage-3.json
+++ b/i686-stage-3.json
@@ -17,5 +17,6 @@
     "os": "none",
     "vendor": "unknown",
     "relocation-model": "static",
-    "features": "+soft-float,-sse,-mmx"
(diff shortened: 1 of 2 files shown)


Freax13:
Given that [0.9.29](https://crates.io/crates/bootloader/0.9.29) is more widely used than [0.11.9](https://crates.io/crates/bootloader/0.11.9) (probably because of [blog_os](https://github.com/phil-opp/blog_os/)?), I think we should backport this fix.

phil-opp (approved):
Thank you!

PR #492, opened 2025-02-04
change rustc-abi in custom targets to x86-softfloat

Freax13:
With the latest nightly, setting "+soft-float" in "features" is only allowed if "rustc-abi" is set to "x86-softfloat".

Fixes a nightly breakage introduced by https://github.com/rust-lang/rust/pull/136146
Related to #491


phil-opp:
Published as v0.9.30

PR #496, opened 2025-02-25
Use `Result::ok`

ChocolateLoverRaj:


uefi/src/main.rs
```diff
@@ -313,10 +313,7 @@ fn load_file_from_disk(
 
     let file_handle_result = root.open(filename, FileMode::Read, FileAttribute::empty());
 
-    let file_handle = match file_handle_result {
-        Err(_) => return None,
-        Ok(handle) => handle,
-    };
+    let file_handle = file_handle_result.ok();
```
Freax13, line 316:
```suggestion
    let file_handle = file_handle_result.ok()?;
```

Freax13 (changes requested):

Freax13 (approved):
Thank you for yor contribution!

LGTM!

PR #500, opened 2025-03-22
remove #[no_mangle] from panic handler

Freax13:
A recent change on rustc nightly broke panic handlers that also have the #[no_mangle] attribute. I don't know whether this is something that they want to support and if they'll fix that, but either way I don't think there's a strong reason to keep this, so let's just remove it.


phil-opp (approved):

phil-opp:
Thanks a lot for looking into this!

phil-opp:
Published as v0.9.31

PR #502, opened 2025-04-01
implement Send+Sync for MemoryRegions

Freax13:
This also makes BootInfo Send+Sync. It's now possible to store BootInfo in a static without resorting to unsafe code.


phil-opp (approved):
Seems reasonable to me, thanks!

PR #509, opened 2025-07-31
Fixes the type of target-c-int-width in target jsons.

OmegaMetor:
In https://github.com/rust-lang/rust/pull/144218, the type of target-c-int-width was changed to an Option<u16>, when previously it could be parsed from a string.
That change caused an error "error: error loading target specification: target-c-int-width: invalid type: string "32", expected u16 at line 13 column 27"
This pr is meant to fix that, by changing the value in the jsons from "32" to 32.


Freax13:
I'll start a new release.

PR #512, opened 2025-07-31
fix target-c-int-width for 0.9.x

### LINKED ISSUE #511: target-c-int-width: invalid type: string "32", expected u16
My custom triple target looks like:

```
{
  "llvm-target": "x86_64-unknown-none",
  "data-layout": "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128",
  "arch": "x86_64",
  "target-endian": "little",
  "target-pointer-width": "64",
  "target-c-int-width": 32,
  "os": "none",
  "executables": true,
  "linker-flavor": "ld.lld",
  "linker": "rust-lld",
  "panic-strategy": "abort",
  "disable-redzone": true,
  "features": "-mmx,-sse,+soft-float",
  "rustc-abi": "x86-softfloat"
}
```

When I run:
`cargo bootimage
`

I get:
`error: error loading target specification: target-c-int-width: invalid type: string "32", expected u16 at line 11 column 28
`

however, the trace of the error takes me to bootloader-0.9.31/x86_64-bootloader.json - where it's occuring as it has `"32"` as the target-c-int-width. I tried updating to the latest bootloader crate but everything breaks on build for some reason so it's unusable. 

This is what I'm running:
```
rustc 1.90.0-nightly (3048886e5 2025-07-30)
binary: rustc
commit-hash: 3048886e59c94470e726ecaaf2add7242510ac11
commit-date: 2025-07-30
host: aarch64-apple-darwin
release: 1.90.0-nightly
LLVM version: 20.1.8
```

Freax13:
Newer rustc versions require this value to be an integer.

For more information see #509.

Closes #511

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
 name = "bootloader"
-version = "0.9.31"
+version = "0.9.32"
 dependencies = [
  "bit_field",
  "fixedvec",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bootloader"
-version = "0.9.31"
+version = "0.9.32"
 authors = ["Philipp Oppermann <dev@phil-opp.com>"]
 license = "MIT/Apache-2.0"
 description = "An experimental pure-Rust x86 bootloader."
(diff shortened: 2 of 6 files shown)


phil-opp (approved):
Thank you!

phil-opp:
Published as `v0.9.32`, sorry for the delay!

I think you should also be able to push the crates.io releases. For the v0.9 branch I just do a `cargo publish --no-verify`.

PR #518, opened 2025-09-03
Fix(v0.9): target-pointer-width field now expects an integer

phil-opp:
See rust-lang/rust#144443


phil-opp:
Published as v0.9.33

PR #519, opened 2025-11-01
Add basic example

peppergrayxyz:
Currently, it is difficult to figure out how all parts play together. There is documentation available, but it needs to be pieced together. To make things easier, it would be great to have a basic kernel that is compiled into a basic OS and booted.

Addition in this PR:
- showcase an OS that boots a minimal kernel using bootloader
- update usage guides
- ~~describe the setup of basic-os with basic-kernel in tutorial.md~~
- ~~use [bashtestmd](https://crates.io/crates/bashtestmd/versions) to create a shell script from the tutorial~~
- ~~run the shell script to generate the basic example~~
- boot the basic example using qemu and show that it starts and prints a message
- ~~add generation of basic example from tutorial to CI~~

This gives newcomers a starting point and can serve as a template for new projects. ~~By generating the basic example from the tutorial and adding this to CI the tutorial and basic example stay consistent and recent.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,15 +43,7 @@ To combine your kernel with `bootloader` and create a bootable disk image, follo
     };
     bootloader_api::entry_point!(kernel_main, config = &CONFIG);
     ```
-- Compile your kernel to an ELF executable by running **`cargo build --target x86_64-unknown-none`**. You might need to run `rustup target add x86_64-unknown-none` before to download precompiled versions of the `core` and `alloc` crates. You can add `x86_64-unknown-none` as default target and add it to your toolchain so that `cargo build` takes care of this.
-    ```toml
-    # .cargo/config.toml
-    [build]
-    target = "x86_64-unknown-none"
-    ```
-    ```sh
-    $ cargo build
-    ```
+- Compile your kernel to an ELF executable by running **`cargo build --target x86_64-unknown-none`**. You might need to run `rustup target add x86_64-unknown-none` before to download precompiled versions of the `core` and `alloc` crates.
 - Thanks to the `entry_point` macro, the compiled executable contains a special section with metadata and the serialized config, which will enable the `bootloader` crate to load it.
 
 #### Creating a bootable image
@@ -61,12 +53,16 @@ To combine your kernel with `bootloader` and create a bootable disk image, follo
     ```sh
     $ cargo init --bin
     ```
-- Define a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html) and add your kernel as a workspace member.
+- Define a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html).
     ```toml
     # in Cargo.toml
     [workspace]
     resolver = "3"
-    members = ["kernel"]
+    members = []
+    ```
+- Add your kernel as a workspace member.
+    ```sh
+    $ cargo new kernel --bin
     ```
 - Enable the workspace to build your kernel:
   - Set up an [artifact dependency](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dependencies) to add your `kernel` crate as a `build-dependency`:
@@ -88,7 +84,7 @@ To combine your kernel with `bootloader` and create a bootable disk image, follo
       channel = "nightly"
       targets = ["x86_64-unknown-none"]
       ```
-  - Alternatively, you can use [`std::process::Command`](https://doc.rust-lang.org/stable/std/process/struct.Command.html) to invoke the build command of your kernel in the `build.rs` script. 
+  - Alternatively, you can use [`std::process::Command`](https://doc.rust-lang.org/stable/std/process/struct.Command.html) to invoke the build command of your kernel in the `build.rs` script.
 - Create a [`build.rs`](https://doc.rust-lang.org/cargo/reference/build-scripts.html) build script in the `os` crate. See our [disk image creation template](docs/create-disk-image.md) for a more detailed example.
   - Obtain the path to the kernel executable. When using an artifact dependency, you can retrieve this path using `std::env::var_os("CARGO_BIN_FILE_MY_KERNEL_my-kernel")`
   - Use `bootloader::UefiBoot` and/or `bootloader::BiosBoot` to create a bootable disk image with your kernel.
--- a/examples/basic/basic-os.md
+++ b/examples/basic/basic-os.md
@@ -35,7 +35,7 @@ The `basic` create combines the kernel with `bootloader`, creates a bootable dis
 - A [cargo build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html) is used to create a bootable disk image ([`build.rs`](build.rs)).
 - `basic` launches the images in QEMU.
 
-See also [basic-kernel.md](kernel/basic-kernel.md) in the `kernel` subdirectory. 
+See also [basic-kernel.md](kernel/basic-kernel.md) in the `kernel` subdirectory.
 
 ## Usage
 
@@ -49,8 +49,8 @@ $ cargo build
 ```
 Run:
 ```sh
-$ cargo run -- bios
+$ cargo run bios
 ```
 ```sh
-$ cargo run -- uefi
+$ cargo run uefi
 ```
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,10 +99,10 @@ jobs:
         run: cargo build
         working-directory: examples/basic
       - name: "Run basic example (bios)"
-        run: cargo run -- bios
+        run: cargo run bios
         working-directory: examples/basic
       - name: "Run basic example (uefi)"
-        run: cargo run -- uefi
+        run: cargo run uefi
         working-directory: examples/basic
 
   fmt:


peppergrayxyz:
Thanks for your Feedback! 

I removed the tutorial and updated usage guides. Please take a look if this is going in the right direction now. Happy to make further adjustments!

examples/basic/kernel/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "kernel"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+bootloader_api = "0.11.12"
+uart_16550 = "0.4.0"
+x86_64 = { version = "0.15.2", features = ["instructions"] }
```
Freax13:
```suggestion
x86_64 = "0.15.2"
```
`instructions` is enabled by default.

examples/basic/kernel/basic-kernel.md
```diff
@@ -0,0 +1,11 @@
...
+
+A minimal kernel to showcase the usage of bootloader.
+
+The kernel:
+- Initalizes a serial port on [COM1](https://wiki.osdev.org/Serial_Ports#Port_Addresses)
+- Dumps the [`boot_info`](/api/src/info.rs) it recevied from `bootloader`
+- Prints a message
+- Exits QEMU with a custom exit code via the `isa-debug-exit` device
+
+Check https://os.phil-opp.com/ for further information on how to write an OS in Rust.
```
Freax13:
```suggestion
```
That tutorial is already very widely known. I'd guess that most people are already familiar with that one and if they aren't I think we also want people to give a chance to stumble upon other tutorials that have been written since.

examples/basic/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+        let code = prebuilt.get_file(Arch::X64, FileType::Code);
+        let vars = prebuilt.get_file(Arch::X64, FileType::Vars);
+
+        cmd.arg("-drive")
+            .arg(format!("format=raw,file={uefi_path}"));
+        cmd.arg("-drive").arg(format!(
+            "if=pflash,format=raw,unit=0,file={},readonly=on",
+            code.display()
+        ));
+        // copy vars and enable rw instead of snapshot if you want to store data (e.g. enrole secure boot keys)
```
Freax13:
```suggestion
        // copy vars and enable rw instead of snapshot if you want to store data (e.g. enroll secure boot keys)
```

examples/basic/basic-os.md
```diff
@@ -0,0 +1,56 @@
...
+    ```toml
+    # rust-toolchain.toml
+    [toolchain]
+    channel = "nightly"
+    targets = ["x86_64-unknown-none"]
+    ```
+
+The `basic` create combines the kernel with `bootloader`, creates a bootable disk image and launches them.
+- A [cargo build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html) is used to create a bootable disk image ([`build.rs`](build.rs)).
+- The images are launched in QEMU by the `main` function ([`src/main.rs`](src/main.rs)).
```
Freax13:
```suggestion
- `basic` launches the images are launched in QEMU.
```

examples/basic/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[dependencies]
+ovmf-prebuilt = "0.2.4"
+
+[workspace]
+resolver = "3"
+members = ["kernel"]
+
+[build-dependencies]
+kernel = { path = "kernel", artifact = "bin", target = "x86_64-unknown-none" }
+bootloader = "0.11.12"
```
Freax13:
```suggestion
[workspace]
resolver = "3"
members = ["kernel"]

[package]
name = "basic"
version = "0.1.0"
edition = "2024"

[dependencies]
ovmf-prebuilt = "0.2.4"

[build-dependencies]
kernel = { path = "kernel", artifact = "bin", target = "x86_64-unknown-none" }
bootloader = "0.11.12"
```

Let separate the sections for the workspace and the top-level crate.

docs/create-disk-image.md
```diff
@@ -5,83 +5,18 @@ The [`bootloader`](https://docs.rs/bootloader/0.11) crate provides simple functi
...
-Now you should be able to use `cargo build` to create a bootable disk image and `cargo run` to run in QEMU. Your kernel is automatically recompiled when it changes. For more advanced usage, you can add command-line arguments to your `main.rs` to e.g. pass additional arguments to QEMU or to copy the disk images to some path to make it easier to find them (e.g. for copying them to an thumb drive).
+Our [basic example](examples/basic/basic-os.md) showcases this setup:
+- [Cargo.toml](/examples/basic/Cargo.toml)
+    - create a workspace & add kernel as member
+    - add kernel as build-dependency
+    - add ovmf-prebuilt for UEFI booting in QEMU
+- [.cargo/config.toml](/examples/basic/Cargo.toml)
+    - enable the unstable artifact-dependencies feature
+- [rust-toolchain.toml](/examples/basic/Cargo.toml)
+    - use experimental features (nightly)
```
Freax13: The rust-toolchain file doesn't enable any experimental features, it just changes the default toolchain.

README.md
```diff
@@ -35,37 +43,57 @@ To make your kernel compatible with `bootloader`:
...
-  [unstable]
-  # enable the unstable artifact-dependencies feature, see
-  # https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dependencies
-  bindeps = true
-  ```
-  Alternatively, you can use [`std::process::Command`](https://doc.rust-lang.org/stable/std/process/struct.Command.html) to invoke the build command of your kernel in the `build.rs` script. 
-- Obtain the path to the kernel executable. When using an artifact dependency, you can retrieve this path using `std::env::var_os("CARGO_BIN_FILE_MY_KERNEL_my-kernel")`
-- Use `bootloader::UefiBoot` and/or `bootloader::BiosBoot` to create a bootable disk image with your kernel.
+    ```sh
+    $ cargo new os --bin
```
Freax13: `cargo new` will create a new directory instead of using the existing top-level directory. IIRC `cargo init` can be used instead. This command also doesn't quite match the description because it doesn't define a workspace.

README.md
```diff
@@ -35,37 +43,57 @@ To make your kernel compatible with `bootloader`:
...
+    ```
+    ```sh
+    $ cargo build
+    ```
 - Thanks to the `entry_point` macro, the compiled executable contains a special section with metadata and the serialized config, which will enable the `bootloader` crate to load it.
 
-### Booting
-
-To combine your kernel with a bootloader and create a bootable disk image, follow these steps:
+#### Setup the workspace
```
Freax13: Only the first three steps are relevant for setting up a workspace.

How about "Creating a bootable image"?

README.md
```diff
@@ -35,37 +43,57 @@ To make your kernel compatible with `bootloader`:
...
+      channel = "nightly"
+      targets = ["x86_64-unknown-none"]
+      ```
+  - Alternatively, you can use [`std::process::Command`](https://doc.rust-lang.org/stable/std/process/struct.Command.html) to invoke the build command of your kernel in the `build.rs` script. 
+- Create a [`build.rs`](https://doc.rust-lang.org/cargo/reference/build-scripts.html) build script in the `os` crate. See our [disk image creation template](docs/create-disk-image.md) for a more detailed example.
+  - Obtain the path to the kernel executable. When using an artifact dependency, you can retrieve this path using `std::env::var_os("CARGO_BIN_FILE_MY_KERNEL_my-kernel")`
+  - Use `bootloader::UefiBoot` and/or `bootloader::BiosBoot` to create a bootable disk image with your kernel.
 - Do something with the bootable disk images in your `main.rs` function. For example, run them with QEMU.
-
-See our [disk image creation template](docs/create-disk-image.md) for a more detailed example.
```
Freax13: Why was this removed?

Freax13 (changes requested):

docs/create-disk-image.md
peppergrayxyz: Updated the wording

README.md
peppergrayxyz: Changed the command and updated the wording

README.md
```diff
@@ -35,34 +43,55 @@ To make your kernel compatible with `bootloader`:
     };
     bootloader_api::entry_point!(kernel_main, config = &CONFIG);
     ```
-- Compile your kernel to an ELF executable by running **`cargo build --target x86_64-unknown-none`**. You might need to run `rustup target add x86_64-unknown-none` before to download precompiled versions of the `core` and `alloc` crates.
+- Compile your kernel to an ELF executable by running **`cargo build --target x86_64-unknown-none`**. You might need to run `rustup target add x86_64-unknown-none` before to download precompiled versions of the `core` and `alloc` crates. You can add `x86_64-unknown-none` as default target and add it to your toolchain so that `cargo build` takes care of this.
+    ```toml
+    # .cargo/config.toml
+    [build]
+    target = "x86_64-unknown-none"
+    ```
```
Freax13:
```suggestion
```
Let's also remove this.

README.md
```diff
@@ -35,34 +43,55 @@ To make your kernel compatible with `bootloader`:
...
-  bindeps = true
-  ```
-  Alternatively, you can use [`std::process::Command`](https://doc.rust-lang.org/stable/std/process/struct.Command.html) to invoke the build command of your kernel in the `build.rs` script. 
-- Obtain the path to the kernel executable. When using an artifact dependency, you can retrieve this path using `std::env::var_os("CARGO_BIN_FILE_MY_KERNEL_my-kernel")`
-- Use `bootloader::UefiBoot` and/or `bootloader::BiosBoot` to create a bootable disk image with your kernel.
+- Create a new `os` crate at the top level
+    ```sh
+    $ cargo init --bin
+    ```
+- Define a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html) and add your kernel as a workspace member.
```
Freax13: If we're specifying the command used to create the top-level crate (`cargo init --bin` we should do the same for the kernel). We could remove the kernel from the members (and just use `members = []`) and then use `cargo new kernel --bin` in the workspace root.

Freax13:
No worries

PR #526, opened 2025-12-23
Code changes to update to 2024 edition

spencer3035:
I decided I would try to make myself useful and update the code to work with the 2024 edition.

The majority of the changes are just specifying unsafe more explicitly. This shouldn't cause any changes in functionality. I tried to have all the formatting changes be in the last commit, but there were a few that made it into the `edition 2021 -> 2024 fixes` commit.

I'm not sure how you would like to handle the upgrade in terms of versioning and whatnot. Let me know if you would like me to make any changes to the PR (or if you don't want to have these changes merged at all).


phil-opp (approved):
Looks good, thank you!

PR #529, opened 2026-01-05
Add llvm-tools to components

the-ssd:
When trying to run the example, bootloader complained about not having llvm-tools, so here is a small patch adding them


phil-opp (approved):
Thanks!

PR #531, opened 2026-01-10
add kernel stack address to BootInfo

Wasabi375:
I'm not sure if a change to `BootInfo` requires a version bump


Freax13 (approved):
LGTM, thanks!

phil-opp:
> > I'm not sure if a change to `BootInfo` requires a version bump
> 
> Last time we didn't treat this as a breaking change, so we should be fine.

The struct is marked as non-exhaustive, so adding fields should not break anybody.

Wasabi375:
Any chance we can get a release anytime soon. I'm happy to create a PR to bump versions and update the changelog. That said it looks like the next version has to be a major version bump as we switched over to 2024 edition.

Freax13:
> Any chance we can get a release anytime soon. I'm happy to create a PR to bump versions and update the changelog.

That'd be greatly appreciated, thanks!

> That said it looks like the next version has to be a major version bump as we switched over to 2024 edition.

I don't think that's a breaking change though, is it? I can't find any documentation about whether this should be a minor or major change, but I can't think of a reason why it would be a breaking change.

PR #533, opened 2026-01-13
Release 0.11.14

Wasabi375:
As discussed [here](https://github.com/rust-osdev/bootloader/pull/531#issuecomment-3739865813) it should be fine for this to be a minor release.

I left some minor fixes(#529, #532) out of the changelog. I hope this is fine. As far as I can tell those are just changes to improve the development environment of this crate and does not affect anyone using the bootloader.


Freax13 (commented):
> I left some minor fixes(#529, #532) out of the changelog. I hope this is fine. As far as I can tell those are just changes to improve the development environment of this crate and does not affect anyone using the bootloader.

Let's put #529 back in. #529 adds something to the example kernel, not the bootloader itself.

PR #537, opened 2026-02-01
Enable json-target-spec

Freax13:
Nightly versions starting with 2026-02-01 require this flag when JSON targets are used.


phil-opp (approved):
Thank you!

Freax13:
Published as v0.9.34.

PR #521, opened 2025-11-27
Fix stage 2 build by moving `.bss` section

phil-opp:
We get `relocation R_386_16 out of range: 74743 is not in [-32768, 65535]` errors with the latest nightlies, as reported in #520. All the failed relocations are from the `.start` section, referencing the `.bss` section. This commit fixes these errors by moving the `.bss` section in between the `.start` and `.text` sections so that it is closer to the `.start` section.

Fixes #520

PR #561, opened 2026-05-09
Fix building with new rust-nightly.

jarovo:
- **Fixes to enable compiling with new rust.**
 after bumping the tempfile to use new deps, there was this linker issue to be fixed:
```
 = note: rust-lld: error: /home/jhenner/projects/bootloader/target/i386-code16-stage-2/stage-2/deps/bootloader_x86_64_bios_stage_2-2a60fabcbb129033.bootloader_x86_64_bios_stage_2.3f4927ae6a5f4601-cgu.0.rcgu.o:(function bootloader_x86_64_bios_stage_2::start: .text._RNvCs5qRQZAVXUQf_30bootloader_x86_64_bios_stage_25start+0x251): relocation R_386_16 out of range: 73531 is not in [-32768, 65535]; references section '.bss._RNvNvXNtCs5qRQZAVXUQf_30bootloader_x86_64_bios_stage_24diskNtB4_10DiskAccessNtB4_4Read10read_exact7TMP_BUF'
 ```
To address this I made disk_buffer a stack variable instead of static variable because the now large `.text` region pushed the `.bss` so much that it didn't fit to the 16bit address range ( the `.text` precedes the `.bss` in the `stage-2-link.ld` file). Another advantage is that this allowed me to remove the unsafe block. I think I was getting some memory corruption when dealing with the references to the static mem region.

- **Enlarge the stack.**
IIRC After moving the `disk_buffer` to the stack the some of the tests `BIOS boot` and then nothing was happening. After changing the ESP to `0x00080000` all the tests are passing. Only the `large_ramdisk` test sometimes fails. It seems to be flaky - many times it passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - uses: r7kamura/rust-problem-matchers@v1.1.0
       - name: "Run `cargo check`"
-        run: cargo check --all-targets --all
+        run: cargo check --all-targets --all --exclude 'test_kernel*'  # Don't fail on panic_handler doubled definition.
       - name: "Check test kernels"
         run: cargo check --all
         working-directory: tests/test_kernels
@@ -123,7 +123,7 @@ jobs:
       - run: cargo --version --verbose
       - uses: Swatinem/rust-cache@v2
       - uses: r7kamura/rust-problem-matchers@v1.1.0
-      - run: cargo clippy --all --all-targets
+      - run: cargo clippy --all --all-targets  --exclude 'test_kernel*' # Don't fail on panic_handler doubled definition.
       - run: cargo clippy --all
         working-directory: tests/test_kernels
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "anyhow"
-version = "1.0.71"
+version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c7d0618f0e0b7e8ff11427422b64564d5fb0be1940354bfe2e0529b18a9d9b8"
+checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "arrayvec"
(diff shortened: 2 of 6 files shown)


Freax13:
> * **Fixes to enable compiling with new rust.**
>       after bumping the tempfile to use new deps, there was this linker issue to be fixed:
> 
> 
> ```
>  = note: rust-lld: error: /home/jhenner/projects/bootloader/target/i386-code16-stage-2/stage-2/deps/bootloader_x86_64_bios_stage_2-2a60fabcbb129033.bootloader_x86_64_bios_stage_2.3f4927ae6a5f4601-cgu.0.rcgu.o:(function bootloader_x86_64_bios_stage_2::start: .text._RNvCs5qRQZAVXUQf_30bootloader_x86_64_bios_stage_25start+0x251): relocation R_386_16 out of range: 73531 is not in [-32768, 65535]; references section '.bss._RNvNvXNtCs5qRQZAVXUQf_30bootloader_x86_64_bios_stage_24diskNtB4_10DiskAccessNtB4_4Read10read_exact7TMP_BUF'
> ```
> 
> To address this I made disk_buffer a stack variable instead of static variable because the now large `.text` region pushed the `.bss` so much that it didn't fit to the 16bit address range ( the `.text` precedes the `.bss` in the `stage-2-link.ld` file). Another advantage is that this allowed me to remove the unsafe block. I think I was getting some stack corruption when dealing with the references to the static mem region.

This was possibly already fixed by #521/#522.

jarovo:
> > * **Fixes to enable compiling with new rust.**
> >   after bumping the tempfile to use new deps, there was this linker issue to be fixed:
> > 
> > ```
> >  = note: rust-lld: error: /home/jhenner/projects/bootloader/target/i386-code16-stage-2/stage-2/deps/bootloader_x86_64_bios_stage_2-2a60fabcbb129033.bootloader_x86_64_bios_stage_2.3f4927ae6a5f4601-cgu.0.rcgu.o:(function bootloader_x86_64_bios_stage_2::start: .text._RNvCs5qRQZAVXUQf_30bootloader_x86_64_bios_stage_25start+0x251): relocation R_386_16 out of range: 73531 is not in [-32768, 65535]; references section '.bss._RNvNvXNtCs5qRQZAVXUQf_30bootloader_x86_64_bios_stage_24diskNtB4_10DiskAccessNtB4_4Read10read_exact7TMP_BUF'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To address this I made disk_buffer a stack variable instead of static variable because the now large `.text` region pushed the `.bss` so much that it didn't fit to the 16bit address range ( the `.text` precedes the `.bss` in the `stage-2-link.ld` file). Another advantage is that this allowed me to remove the unsafe block. I think I was getting some stack corruption when dealing with the references to the static mem region.
> 
> This was possibly already fixed by #521/#522.

Yes, it seems like got fixed before me noticing it.

May I ask you to leave a comment on the idea of making the `disk_buffer` on stack in stage-2 and setting the base of the stack to `0x00080000` in stage-3 to make it fit the kernel while removing the unsafe section? Perhaps I am missing something, but I think we don't need to preserve the stack between the stages. It is something I would like to check before I un-mark this as Draft.

Freax13:
It's an interesting idea, but one that makes me a bit nervous. rustc isn't always very good with stack usage and this change would get us more than halfway to the limit. I've seen the compiler duplicate individual stack variables when they're being moved around (and double the required stack memory) before and if that were to happen here that'd exceed the stack size. I'd hope it wouldn't do that here, but I'd feel more comfortable if we didn't leave this up to chance.

PR #563, opened 2026-05-10
update to tempfile 3.27.0

Freax13:
Previously, tempfile depended on an old rustix version that no longer builds on the latest nightly compiler. Bumping the tempfile crate also bumps rustix.

Supersedes #561

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - uses: r7kamura/rust-problem-matchers@v1.1.0
       - name: "Run `cargo check`"
-        run: cargo check --all-targets --all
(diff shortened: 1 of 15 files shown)


Freax13:
@jarovo I saw that you also made changes in #561 to move the disk buffer around. I'm not quite sure why that's needed. Is there a reason to do this?

jarovo:
> @jarovo I saw that you also made changes in #561 to move the disk buffer around. I'm not quite sure why that's needed. Is there a reason to do this?

Thanks for looking at the code and asking.  I've just updated the PR description to specify the reasons for the changes.

phil-opp (approved):
Looks good, thank you!

Freax13:
I don't *think* a release is necessary for this, it should be enough for user to just run `cargo update`. Please let me know if I'm wrong.

PR #565, opened 2026-06-14
deps: bump uart_16550 to 0.6.0

phip1611:
Let's use the latest and greatest version

### DIFF
--- a/common/src/serial.rs
+++ b/common/src/serial.rs
@@ -1,7 +1,10 @@
 use core::fmt;
+use uart_16550::backend::PioBackend;
 
+// TODO this type can be replaced with Uart16550Tty but using it currently panics
+// in the new constructor.
 pub struct SerialPort {
-    port: uart_16550::SerialPort,
+    port: uart_16550::Uart16550<PioBackend>,
 }
 
 impl SerialPort {
@@ -9,8 +12,10 @@ impl SerialPort {
     ///
     /// unsafe because this function must only be called once
     pub unsafe fn init() -> Self {
-        let mut port = unsafe { uart_16550::SerialPort::new(0x3F8) };
-        port.init();
+        let mut port =
+            unsafe { uart_16550::Uart16550::new_port(0x3F8) }.expect("should be valid port");
+        port.init(uart_16550::Config::default())
+            .expect("should init successfully");
         Self { port }
     }
 }
@@ -19,8 +24,8 @@ impl fmt::Write for SerialPort {
     fn write_str(&mut self, s: &str) -> fmt::Result {
         for char in s.bytes() {
             match char {
-                b'\n' => self.port.write_str("\r\n").unwrap(),
-                byte => self.port.send(byte),
+                b'\n' => self.port.send_bytes_exact(b"\r\n"),
+                byte => self.port.send_bytes_exact(&[byte]),
             }
         }
         Ok(())
--- a/tests/test_kernels/min_stack/src/bin/basic_boot.rs
+++ b/tests/test_kernels/min_stack/src/bin/basic_boot.rs
@@ -7,7 +7,7 @@ use test_kernel_min_stack::{QemuExitCode, exit_qemu, serial};
 
 const BOOTLOADER_CONFIG: BootloaderConfig = {
     let mut config = BootloaderConfig::new_default();
-    config.kernel_stack_size = 3000;
+    config.kernel_stack_size = 4000;
     config
 };
 entry_point!(kernel_main, config = &BOOTLOADER_CONFIG);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,7 +138,7 @@ dependencies = [
  "rsdp",
  "serde-json-core",
  "usize_conversions",
- "x86_64 0.15.2",
+ "x86_64",
 ]
 
 [[package]]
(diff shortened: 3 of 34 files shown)


tests/test_kernels/min_stack/src/bin/basic_boot.rs
```diff
@@ -7,7 +7,7 @@ use test_kernel_min_stack::{QemuExitCode, exit_qemu, serial};
 
 const BOOTLOADER_CONFIG: BootloaderConfig = {
     let mut config = BootloaderConfig::new_default();
-    config.kernel_stack_size = 3000;
+    config.kernel_stack_size = 3032;
```
phip1611: @phil-opp @Freax13 please take a look. I do not know enough about the current test structure. This change is in a dedicated commit. WDYT?

common/src/serial.rs
```diff
@@ -1,16 +1,19 @@
 use core::fmt;
+use uart_16550::backend::PioBackend;
 
 pub struct SerialPort {
```
Freax13, line 6: Should we replace `SerialPort` with `Uart16550Tty`? Both implement `core::fmt::Write`, so I don't see much of a reason to keep `SerialPort` around. `Uart16550Tty` also does the `\n` to `\r\n` conversion `SerialPort` does.

tests/test_kernels/min_stack/src/bin/basic_boot.rs
Freax13: Judging by the generated assembly, my guess is that most of the stack usage seems to come from `Result<T, Uart16550TtyError>`. `Uart16550TtyError` is quite big (36 bytes). On lower optimization levels, the compiler isn't very smart about this and allocates a variable slot on the stack for every function call that returns `Result<T, Uart16550TtyError>`.

Let's add some safety margin to make sure this doesn't get tripped by future small regressions. How about 4000?

Freax13 (commented):

common/src/serial.rs
phip1611: I tried but for some reason on of these checks fails:

image


This surprises me becaue in the integration test of uart_16550 it works in QEMU. I don't have time to investigate for now. Can we move forward?

I've added a TODO comment in the code

Freax13: Looks to me like it's spinning somewhere. I _think_ it's spinning in `receive_bytes_exact`.

Freax13 (approved):
LGTM, thanks!

common/src/serial.rs
phip1611: could you help me investigating please? I'd love to understand why it fails

Freax13: > could you help me investigating please? I'd love to understand why it fails
> 

I tried looking into this yesterday. I think it's some kind of race condition in the self-test code. I can't trigger the issue reliably though. Adding artificial system load seems to make it easier to reproduce, but it's still not completely reliable.
For some reason, I had a harder time reproducing the error with the latest HEAD for uart_16550. I'm not sure why.
I might be able to take another look at this later today.
I think my next step is going to be looking at the serial port implementation in QEMU to see if there's anything that could explain the race condition (maybe asynchronous processing?).

phip1611: thanks a lot for looking into this! I have to say that this surprises me. I tested the lib a lot in QEMU, Cloud Hypervisor but also real hardware

> I might be able to take another look at this later today.

that would be amazing!

Freax13: I don't know about you, but I've only seen this happen when booting with UEFI. Maybe UEFI's initialization and use of the serial port before the bootloader somehow causes issues?

phip1611: Interesting - the UEFI console takes ownership of the serial device by default. One must therefore disconnect the UEFI console from using the serial device (as long as boot services are active but one wants to use the serial device). I think this is the problem.

The forced disconnct is also fairly easy

Freax13: > The forced disconnct is also fairly easy

How would one do that?

---

Here's something interesting: Wrapping the call to `Uart16550Tty::new_port` in `x86_64::instructions::interrupts::without_interrupts` fixes the issue. Disabling interrupts in the tty config (`config.interrupts = IER::empty();`) does _not_ work however.

phip1611: > > The forced disconnct is also fairly easy
> 
> How would one do that?
> 

https://github.com/phip1611/uefi-serial-chat/blob/47cc3961ab047d698790dc48131799aa10c18c65/src/main.rs#L104

Regarding interrupts:thanks! That's something I can use for my testing and debugging

phip1611: BTW I'm preparing an update of the uefi dependency

phil-opp:
I think this caused a regression on real hardware (see #573) because of https://github.com/rust-osdev/uart_16550/issues/65.

Freax13:
Should we release a workaround until a fix is available?

phil-opp:
I would be in favor. I see two options: either revert this PR for now or gate the serial printing on `check_connected`. The latter should fix the hang, but would skip serial output on typical TX/RX/GND serial adapters...

Freax13:
> I would be in favor. I see two options: either revert this PR for now or gate the serial printing on `check_connected`. The latter should fix the hang, but would skip serial output on typical TX/RX/GND serial adapters...

Let's go with the former.

common/src/serial.rs
phip1611: > Here's something interesting: Wrapping the call to Uart16550Tty::new_port in x86_64::instructions::interrupts::without_interrupts fixes the issue. Disabling interrupts in the tty config (config.interrupts = IER::empty();) does not work however.

Found the issue and fixed/documented everything. v0.8.0 will contain the fix. More info: https://github.com/rust-osdev/uart_16550/pull/69

PR #566, opened 2026-06-16
uefi: bump from 0.20 to 0.38

phip1611:
Bump the outdated uefi code to a much newer `uefi` version + misc code improvements

### DIFF
--- a/uefi/src/main.rs
+++ b/uefi/src/main.rs
@@ -242,8 +242,8 @@ fn allocate_loader_data(size: usize) -> &'static mut [u8] {
         ((size - 1) / 4096) + 1,
     )
     .expect("Failed to allocate memory for the file");
-    unsafe { ptr::write_bytes(ptr.as_mut(), 0, size) };
-    unsafe { slice::from_raw_parts_mut(ptr.as_mut(), size) }
+    unsafe { ptr::write_bytes(ptr.as_ptr(), 0, size) };
+    unsafe { slice::from_raw_parts_mut(ptr.as_ptr(), size) }
 }
 
 /// Opens PXE Base Code for the device path that loaded this image.
--- a/uefi/src/main.rs
+++ b/uefi/src/main.rs
@@ -8,30 +8,22 @@ use bootloader_boot_config::BootConfig;
 use bootloader_x86_64_common::{
     Kernel, RawFrameBufferInfo, SystemInfo, legacy_memory_region::LegacyFrameAllocator,
 };
-use core::{
-    cell::UnsafeCell,
-    ops::{Deref, DerefMut},
-    ptr, slice,
-};
+use core::net::Ipv4Addr;
+use core::{ptr, slice};
+use uefi::mem::memory_map::{MemoryMap, MemoryMapMut};
+use uefi::table::cfg::ConfigTableEntry;
 use uefi::{
-    CStr8, CStr16,
-    prelude::{Boot, Handle, Status, SystemTable, entry},
+    CStr8, CStr16, boot,
+    boot::{AllocateType, MemoryType},
+    cstr8, cstr16,
+    prelude::{Status, entry},
     proto::{
         ProtocolPointer,
         console::gop::{GraphicsOutput, PixelFormat},
         device_path::DevicePath,
         loaded_image::LoadedImage,
-        media::{
-            file::{File, FileAttribute, FileInfo, FileMode},
-            fs::SimpleFileSystem,
-        },
-        network::{
-            IpAddress,
-            pxe::{BaseCode, DhcpV4Packet},
-        },
-    },
-    table::boot::{
-        AllocateType, MemoryType, OpenProtocolAttributes, OpenProtocolParams, ScopedProtocol,
+        media::file::{File, FileAttribute, FileInfo, FileMode},
+        network::pxe::{BaseCode, DhcpV4Packet},
     },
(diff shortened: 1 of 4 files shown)


phip1611:
what do we do about the semver CI issue?

phil-opp:
The semver CI job fails also on main. `cargo-semver-checks` is usually not updated directly when the metadata format changes, so we have to wait a bit until it is supported. Safe to ignore for now

uefi/src/main.rs
```diff
@@ -324,62 +202,70 @@ fn load_file_from_disk(
...
+        AllocateType::AnyPages,
+        MemoryType::LOADER_DATA,
+        ((size - 1) / 4096) + 1,
+    )
+    .expect("Failed to allocate memory for the file");
+    unsafe { ptr::write_bytes(ptr.as_mut(), 0, size) };
+    unsafe { slice::from_raw_parts_mut(ptr.as_mut(), size) }
+}
+
+fn open_pxe_base_code() -> Option<boot::ScopedProtocol<BaseCode>> {
```
Freax13, line 253: This is subtly changing the behavior isn't it? Previously, we opened the BaseCode handle that was used to load the base image, but now we're looking at all BaseCode handles, right? If a system has more than one network interface (e.g. a wireless one and a wired one), this could change which interface is used to load the files, right?

Freax13 (commented):
Thanks, this is great work!

Sorry for the delay.

uefi/src/main.rs
phip1611: oh, that was a wonderful catch! I didn't thought about this. Fixed

phip1611: I've also conducted a "critical LLM review" to check behavioral changes - seems to be in good shape now!

phip1611:
The CI failures seem to be unrealated @Freax13 - probably caused by the new clippy beta/nightly update.


PSA: I'll be on vacation for the rest of the month

phil-opp:
I don't think that the CI failures are spurious. I found this in the logs of the failed `Check` job:

```
[lock_api 0.4.10] error[E0463]: can't find crate for `std`
  [lock_api 0.4.10]   |
  [lock_api 0.4.10]   = note: the `x86_64-unknown-uefi` target may not be installed
  [lock_api 0.4.10]   = help: consider downloading the target with `rustup target add x86_64-unknown-uefi`
  [lock_api 0.4.10]   = help: consider building the standard library from source with `cargo build -Zbuild-std`
  [lock_api 0.4.10] 
  [lock_api 0.4.10] error: aborting due to 1 previous error
  [lock_api 0.4.10] 
  [lock_api 0.4.10] For more information about this error, try `rustc --explain E0463`.
  [lock_api 0.4.10] error[E0463]: can't find crate for `core`
  [lock_api 0.4.10]   |
  [lock_api 0.4.10]   = note: the `x86_64-unknown-uefi` target may not be installed
  [lock_api 0.4.10]   = help: consider downloading the target with `rustup target add x86_64-unknown-uefi`
  [lock_api 0.4.10]   = help: consider building the standard library from source with `cargo build -Zbuild-std`
  [lock_api 0.4.10] 
  [lock_api 0.4.10] error: aborting due to 1 previous error
  [lock_api 0.4.10] 
  [lock_api 0.4.10] For more information about this error, try `rustc --explain E0463`.
  [lock_api 0.4.10] warning: autocfg could not probe for `std`
```

So something is trying to bring in a `std` dependency, which doesn't work for this bare-metal project.

I'll look into it, have a good vacation @phip1611!


_edit:_ Ah, that's just a `autocfg` thing, not an actual error.

phil-opp:
Ok, the actual error is:

```
error: error loading target specification: rustc-abi: invalid rustc abi: 'x86-softfloat'. allowed values: 'x86-sse2', 'powerpc-spe', 'softfloat' at line 23 column 1
```

This is fixed by #569 , so a merge/rebase should be enough to fix it.

uefi/src/main.rs
```diff
@@ -324,62 +205,66 @@ fn load_file_from_disk(
...
 
+fn allocate_loader_data(size: usize) -> &'static mut [u8] {
+    let mut ptr = boot::allocate_pages(
+        AllocateType::AnyPages,
+        MemoryType::LOADER_DATA,
+        ((size - 1) / 4096) + 1,
+    )
+    .expect("Failed to allocate memory for the file");
+    unsafe { ptr::write_bytes(ptr.as_mut(), 0, size) };
+    unsafe { slice::from_raw_parts_mut(ptr.as_mut(), size) }
```
phil-opp: The `as_mut` calls create an intermediate `&mut u8` reference before they're implicitly cast to a `*mut u8` pointer. This is problematic under Rust aliasing/provenance model since a `&mut u8` only has access to a single `u8`, but we write/read a whole `[u8]` array through it.

So using the pointer directly through `as_ptr` would be better:

```suggestion
    unsafe { ptr::write_bytes(ptr.as_ptr(), 0, size) };
    unsafe { slice::from_raw_parts_mut(ptr.as_ptr(), size) }
```

phil-opp (approved):

uefi/src/main.rs
Freax13: > I've also conducted a "critical LLM review" to check behavioral changes - seems to be in good shape now!

how'd you do that?

PR #567, opened 2026-06-25
janked dependencies: bump critical-section, atomic-polyfill, heapless

Wasabi375:
We currently rely on dependencies that are janked on crates.io. This causes warning when the bootlader is installed using the locked flag as is done in build.rs.

This PR bumps critical-section and atomic-polyfill. Also heapless gets bumped as it requires an exact version of atomic-polyfill so it needs to be bumped for cargo to accept a newer version of atomic-polyfill.

Not sure what the best practices are here. It looks like updates to Cargo.lock in this repo seem to only change versions of specific crates when needed. Maybe a clean `cargo update` should be done at some point, considering the bootloader is installed using `--locked`.


Freax13 (approved):

PR #571, opened 2026-07-20
Fix workspace.exclude in Cargo.toml

Wasabi375:
Exclude does not support globs. Instead it uses `starts_with`.

See: https://github.com/rust-lang/cargo/issues/6009


Freax13 (approved):
Thanks!

PR #577, opened 2026-07-27
Make release script more robust and send user agent

phil-opp:
The data access policy of crates.io requires a `user-agent` header that identifies the application. Otherwise it may block the request. See https://crates.io/data-access for details.

This commit adds the required user-agent header and also fixes a related bug that misclassified the error response as 'release does not exist yet'. We now only create a new release if we're sure that the release does not exist yet. Errors are raised and fail the CI job so that we can investigate them.

---

Discovered in https://github.com/rust-osdev/bootloader/actions/runs/30249893235/job/89925267279 

```
gh: Validation Failed (HTTP 422)
{"message":"Validation Failed","errors":[{"resource":"Release","code":"already_exists","field":"tag_name"}],"documentation_url":"https://docs.github.com/rest/releases/releases#create-a-release","status":"422"}Detected crate version 0.11.16
Could not find version 0.11.16 on crates.io; creating a new release
```

The API returned an error here and the script misclassified it as 'no release exists' and tried to publish.

### DIFF
--- a/.github/workflows/trigger-release/trigger-release.py
+++ b/.github/workflows/trigger-release/trigger-release.py
@@ -1,3 +1,4 @@
+import sys
 import toml
 import requests
 import subprocess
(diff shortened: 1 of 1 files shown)


Freax13 (approved):
LGTM, thanks!

PR #580, opened 2026-08-15
deps: bump uart_16550 to 0.8.0 + fix UEFI weirdness

phip1611:
This PR upgrades the bootloader to `uart_16550` 0.8.0, bringing its modernized, real-hardware-focused UART driver into the project now that the upstream regression is resolved.

Alongside the API migration, the UEFI path explicitly disconnects the UEFI console from the serial device before initializing the bootloader logger. This gives the logger exclusive ownership of the UART, preventing conflicting device access and avoiding UEFI console output being duplicated on the serial port. 

Some context for some weirdness that @Freax13 found in an earlier PR (https://github.com/rust-osdev/bootloader/pull/565#discussion_r3790095260):

- By default I enabled some interrupts in the Config of the Uart16550
- As UEFI was driving the device, interrupts were firing before init() was finished
- Now:
  - UEFI does no longer take ownership over the device (good)
  - even if it would, the default config now doesn't activate any interrupts of the device (also good, as not needed here)

## Testing

I tested the crate on real hardware (https://github.com/rust-osdev/uart_16550/pull/71) but not the bootloader crate. If you give me instructions for how to test it best without much effort, it would be highly appreciated.


Freax13:
Is there a reason this PR was marked as a draft?

phip1611:
> Is there a reason this PR was marked as a draft?

yup, sorry, I'm doing some final testing!

phip1611:
@Freax13 rdy for review now!

Freax13:
> If you give me instructions for how to test it best without much effort, it would be highly appreciated.

A few years ago I made a small kernel specifically for testing on real hardware: https://github.com/Freax13/test-kernel.

I created a branch that replaces the bootloader dependency with this PR: https://github.com/Freax13/test-kernel/tree/test-uart (that's not quite what I used to test this PR, but it should be very close).

I just tested an UEFI image based on this PR on a bunch of different devices (1x AMD laptop, 1x Intel laptop, 2x AMD desktops, 1x AMD server, 1x Intel server, 2x Intel NUCs). I didn't encounter any issues on any of these devices. Two of these had a serial port (1x AMD server and 1x Intel NUCs) and the serial output also worked flawlessly.

Let me know if you also want to do a bit more testing or if you're happy with the current level of testing.

uefi/src/main.rs
```diff
@@ -395,6 +397,15 @@ fn init_logger(config: &BootConfig) -> Option<RawFrameBufferInfo> {
...
 
+    // Disconnect the UEFI console from the serial device.
+    // Our own logger takes full ownership of the serial device.
+    {
+        if let Ok(handles) = boot::locate_handle_buffer(SearchType::ByProtocol(&Serial::GUID)) {
+            for handle in handles.iter() {
+                let _ = boot::disconnect_controller(*handle, None, None);
+            }
+        };
+    }
```
Freax13:
```suggestion
    if let Ok(handles) = boot::locate_handle_buffer(SearchType::ByProtocol(&Serial::GUID)) {
        for handle in handles.iter() {
            let _ = boot::disconnect_controller(*handle, None, None);
        }
    }
```

phip1611: thanks!

phip1611:
> Let me know if you also want to do a bit more testing or if you're happy with the current level of testing.

From my side, I'm confident this is the right approach.

Freax13 (approved):
LGTM, thanks! :shipit:

